PR #2, opened 2019-03-03
Require Laravel 5.5 or greater

matthewhall-ca:
Update the composer.json file to require Laravel 5.5 or greater


taka-oyama:
Hi, thank you for the fix!

PR #22, opened 2020-11-20
Add support of foreign keys in Blueprint

dkop:
Foreign keys are supported now by Google Spanner


taka-oyama:
Hi! Thanks for the fix.

Issue #37, opened 2021-10-05
Session not found Exception


mgcostaParedes:
Hello after trying to setup your package inside a PHP container, and after installing all the requirements needed for this package, i found a constant error using it, for example, if i create a simple route with the following code:

```
$results = DB::connection('spanner')->table('users')->get();
dd($results);
```

It works perfectly on the first calling, but after that, it throws the following exception:

![Screenshot from 2021-10-05 15-49-38](https://user-images.githubusercontent.com/16031444/136046950-875266fe-27c2-416e-9708-1f670e75fda9.png)
:

and it will just throw it, until i delete the spanner session files, inside (storage/framework/cache/spanner), to force on creating a new session, looks like the package cannot decode properly the cached session?

taka-oyama:
I use this driver daily with the latest and have never seen this error.

Can you check to see that it's actually being written correctly?

Also, you can also override `Colopl/Spanner/SpannerServiceProvider` and use some other adapter (like APCu) to see if the session itself is actually valid.

taka-oyama:
Hi, this was recently fixed with v4.1.0 or 3.10.0 (if you're still on laravel 8 or below).

PR #39, opened 2022-05-15
feat: handle 'session not found' errors

oprudkyi:
fixes https://github.com/colopl/laravel-spanner/issues/37

There are a few cases when a 'Session not found' error can
[happen](https://cloud.google.com/spanner/docs/sessions#handle_deleted_sessions):

 - Scripts that idle too long - for example, a [Laravel queue worker](https://laravel.com/docs/9.x/queues#the-queue-work-command)
or anything that doesn't call Spanner frequently enough (more than once an hour).
 - The session is more than 28 days old.
 - Some random flukes on Google's side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,18 +248,18 @@ or anything that doesn't call Spanner frequently enough (more than once an hour)
 
 The errors can be handled by one of the supported modes:
 
-- **MAINTAIN_SESSION_POOL** - When the 'session not found' error is encountered, the library tries to disconnect,
+- **MAINTAIN_SESSION_POOL** (default) - When the 'session not found' error is encountered, the library tries to disconnect,
 [maintain a session pool](https://github.com/googleapis/google-cloud-php/blob/077810260b58f5de8a3bbdfd999a5e9a48f71a7f/Spanner/src/Session/CacheSessionPool.php#L864)
 (to remove outdated sessions), reconnect, and then try querying again.
 ```php
         'spanner' => [
             'driver' => 'spanner',
         ...
-   	        'sessionNotFoundErrorMode' => 'MAINTAIN_SESSION_POOL',
+            'sessionNotFoundErrorMode' => 'MAINTAIN_SESSION_POOL',
         ]
 ```
 
-- **CLEAR_SESSION_POOL** (default) - The **MAINTAIN_SESSION_POOL** mode is tried first. If the error still happens, then
+- **CLEAR_SESSION_POOL** - The **MAINTAIN_SESSION_POOL** mode is tried first. If the error still happens, then
 the [clearing of the session pool](https://github.com/googleapis/google-cloud-php/blob/077810260b58f5de8a3bbdfd999a5e9a48f71a7f/Spanner/src/Session/CacheSessionPool.php#L465)
 is enforced and the query is tried once again.
 As a consequence of session pool clearing, all processes that share the current session pool will be forced
@@ -268,16 +268,16 @@ to use the new session on the next call. The mode is enabled by default, but you
         'spanner' => [
             'driver' => 'spanner',
         ...
-       	    'sessionNotFoundErrorMode' => 'CLEAR_SESSION_POOL'
+            'sessionNotFoundErrorMode' => 'CLEAR_SESSION_POOL'
         ]
 ```
 
-- none - The QueryException is raised and the client code is free to handle it by itself.:
+- **THROW_EXCEPTION** - The QueryException is raised and the client code is free to handle it by itself.:
 ```php
         'spanner' => [
             'driver' => 'spanner',
         ...
-   	        'sessionNotFoundErrorMode' => false,
+            'sessionNotFoundErrorMode' => 'THROW_EXCEPTION',
         ]
 ```
 
--- a/README.md
+++ b/README.md
@@ -236,6 +236,68 @@ In order to improve the performance of the first connection per request, we use 
 By default, laravel-spanner uses [Filesystem Cache Adapter](https://symfony.com/doc/current/components/cache/adapters/filesystem_adapter.html) as the caching pool. If you want to use your own caching pool, you can extend ServiceProvider and inject it into the constructor of `Colopl\Spanner\Connection`.
 
 
+
+### 'Session not found' exception handling
+There are a few cases when a 'Session not found' error can
+[happen](https://cloud.google.com/spanner/docs/sessions#handle_deleted_sessions):
+
+ - Scripts that idle too long - for example, a [Laravel queue worker](https://laravel.com/docs/9.x/queues#the-queue-work-command)
+or anything that doesn't call Spanner frequently enough (more than once an hour).
+ - The session is more than 28 days old.
+ - Some random flukes on Google's side.
+
+The errors can be handled by one of the supported modes:
+
+- **MAINTAIN_SESSION_POOL** - When the 'session not found' error is encountered, the library tries to disconnect,
+[maintain a session pool](https://github.com/googleapis/google-cloud-php/blob/077810260b58f5de8a3bbdfd999a5e9a48f71a7f/Spanner/src/Session/CacheSessionPool.php#L864)
+(to remove outdated sessions), reconnect, and then try querying again.
+```php
+        'spanner' => [
+            'driver' => 'spanner',
+        ...
+   	        'sessionNotFoundErrorMode' => 'MAINTAIN_SESSION_POOL',
+        ]
+```
+
+- **CLEAR_SESSION_POOL** (default) - The **MAINTAIN_SESSION_POOL** mode is tried first. If the error still happens, then
+the [clearing of the session pool](https://github.com/googleapis/google-cloud-php/blob/077810260b58f5de8a3bbdfd999a5e9a48f71a7f/Spanner/src/Session/CacheSessionPool.php#L465)
+is enforced and the query is tried once again.
+As a consequence of session pool clearing, all processes that share the current session pool will be forced
+to use the new session on the next call. The mode is enabled by default, but you can enable it explicitly via congifuration:
+```php
+        'spanner' => [
+            'driver' => 'spanner',
+        ...
+       	    'sessionNotFoundErrorMode' => 'CLEAR_SESSION_POOL'
+        ]
+```
+
+- none - The QueryException is raised and the client code is free to handle it by itself.:
+```php
+        'spanner' => [
+            'driver' => 'spanner',
+        ...
+   	        'sessionNotFoundErrorMode' => false,
+        ]
+```
+
+Please note, that [`getDatabaseContext()->execute(...)->rows()`](https://github.com/googleapis/google-cloud-php/blob/077810260b58f5de8a3bbdfd999a5e9a48f71a7f/Spanner/src/Result.php#L175)
+returns a `/Generator` object, which only accesses Spanner when iterated. That affects `cursor()`
+and `cursorWithTimestampBound()` functions and many low-level calls. So you might still
+get `Google\Cloud\Core\Exception\NotFoundException` when trying to resolve cursor.
+To avoid that, please run cursor* functions inside
+[explicit transactions](#transactions) so statements will repeat on error.
+
+```php
+$conn->transaction(function () use ($conn) {
+    $cursor = $conn->cursor('SELECT ...');
+
+    foearch ($cursor as $value) { ...
+});
+```
+
+
+
 ### Laravel Tinker
 You can use [Laravel Tinker](https://github.com/laravel/tinker) with commands such as `php artisan tinker`.
 But your session may hang when accessing Cloud Spanner. This is known gRPC issue that occurs when PHP forks a process.
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ The errors can be handled by one of the supported modes:
 - **MAINTAIN_SESSION_POOL** (default) - When the 'session not found' error is encountered, the library tries to disconnect,
 [maintain a session pool](https://github.com/googleapis/google-cloud-php/blob/077810260b58f5de8a3bbdfd999a5e9a48f71a7f/Spanner/src/Session/CacheSessionPool.php#L864)
 (to remove outdated sessions), reconnect, and then try querying again.
+The mode is enabled by default, but you can enable it explicitly via congifuration:
 ```php
         'spanner' => [
             'driver' => 'spanner',
@@ -263,7 +264,7 @@ The errors can be handled by one of the supported modes:
 the [clearing of the session pool](https://github.com/googleapis/google-cloud-php/blob/077810260b58f5de8a3bbdfd999a5e9a48f71a7f/Spanner/src/Session/CacheSessionPool.php#L465)
 is enforced and the query is tried once again.
 As a consequence of session pool clearing, all processes that share the current session pool will be forced
-to use the new session on the next call. The mode is enabled by default, but you can enable it explicitly via congifuration:
+to use the new session on the next call.
 ```php
         'spanner' => [
             'driver' => 'spanner',
--- a/src/Colopl/Spanner/Concerns/ManagesSessionPool.php
+++ b/src/Colopl/Spanner/Concerns/ManagesSessionPool.php
@@ -81,7 +81,8 @@ trait ManagesSessionPool
     {
         $databaseName = $this->getSpannerDatabase()->name();
         $response = (new ProtobufSpannerClient())->listSessions($databaseName);
-        return collect($response->iterateAllElements())->map(function (ProtobufSpannerSession $session) {
+        return collect($response->iterateAllElements())->map(function ($session) {
+            assert($session instanceof ProtobufSpannerSession);
             return new Session($session);
         });
     }
--- a/src/Colopl/Spanner/Concerns/ManagesStaleReads.php
+++ b/src/Colopl/Spanner/Concerns/ManagesStaleReads.php
@@ -56,7 +56,7 @@ trait ManagesStaleReads
      */
     public function selectWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): array
     {
-        return $this->sessionNotFoundWrapper(function () use ($query, $bindings, $timestampBound) {
+        return $this->withSessionNotFoundHandling(function () use ($query, $bindings, $timestampBound) {
             return iterator_to_array($this->cursorWithTimestampBound($query, $bindings, $timestampBound));
         });
     }
@@ -70,7 +70,7 @@ trait ManagesStaleReads
      */
     public function selectOneWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): ?array
     {
-        $result = $this->sessionNotFoundWrapper(function () use ($query, $bindings, $timestampBound) {
+        $result = $this->withSessionNotFoundHandling(function () use ($query, $bindings, $timestampBound) {
             return $this->cursorWithTimestampBound($query, $bindings, $timestampBound)->current();
         });
         assert(is_null($result) || is_array($result));
--- a/src/Colopl/Spanner/Concerns/ManagesStaleReads.php
+++ b/src/Colopl/Spanner/Concerns/ManagesStaleReads.php
@@ -56,7 +56,9 @@ trait ManagesStaleReads
      */
     public function selectWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): array
     {
-        return iterator_to_array($this->cursorWithTimestampBound($query, $bindings, $timestampBound));
+        return $this->sessionNotFoundWrapper(function () use ($query, $bindings, $timestampBound) {
+            return iterator_to_array($this->cursorWithTimestampBound($query, $bindings, $timestampBound));
+        });
     }
 
     /**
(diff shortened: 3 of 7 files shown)


src/Colopl/Spanner/Concerns/ManagesSessionPool.php
```diff
@@ -81,7 +81,8 @@ public function listSessions(): Collection
     {
         $databaseName = $this->getSpannerDatabase()->name();
         $response = (new ProtobufSpannerClient())->listSessions($databaseName);
-        return collect($response->iterateAllElements())->map(function (ProtobufSpannerSession $session) {
+        return collect($response->iterateAllElements())->map(function ($session) {
+            assert($session instanceof ProtobufSpannerSession);
```
taka-oyama, line 85: Why was this changed to an assertion?

src/Colopl/Spanner/Concerns/ManagesStaleReads.php
```diff
@@ -56,7 +56,9 @@ public function cursorWithTimestampBound($query, $bindings = [], TimestampBoundI
      */
     public function selectWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): array
     {
-        return iterator_to_array($this->cursorWithTimestampBound($query, $bindings, $timestampBound));
+        return $this->sessionNotFoundWrapper(function () use ($query, $bindings, $timestampBound) {
+			return iterator_to_array($this->cursorWithTimestampBound($query, $bindings, $timestampBound));
```
taka-oyama, line 60: Please use 4 spaces instead of tabs.

src/Colopl/Spanner/Concerns/ManagesStaleReads.php
```diff
@@ -68,7 +70,11 @@ public function selectWithTimestampBound($query, $bindings = [], TimestampBoundI
      */
     public function selectOneWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): ?array
     {
-        return $this->cursorWithTimestampBound($query, $bindings, $timestampBound)->current();
+        $result = $this->sessionNotFoundWrapper(function () use ($query, $bindings, $timestampBound) {
+			return $this->cursorWithTimestampBound($query, $bindings, $timestampBound)->current();
+		});
+        assert(is_null($result) || is_array($result));
```
taka-oyama, line 76: Why is there an assertion here?

src/Colopl/Spanner/Concerns/ManagesTransactions.php
```diff
@@ -158,7 +161,14 @@ protected function performRollBack($toLevel)
 
         if ($this->currentTransaction !== null) {
             if ($this->currentTransaction->state() === Transaction::STATE_ACTIVE) {
-                $this->currentTransaction->rollBack();
+                try {
+                    $this->currentTransaction->rollBack();
+                } catch (NotFoundException $e) {
+                    // ignore session not found error
```
taka-oyama, line 167: Wouldn't ignoring the error cause unexpected behaviour upstream?

src/Colopl/Spanner/Connection.php
```diff
@@ -443,4 +462,83 @@ protected function runQueryCallback($query, $bindings, Closure $callback)
...
+            throw new InvalidArgumentException("Unsupported sessionNotFoundErrorMode [{$handlerMode}].");
+        }
+        try {
+            return $callback();
+        } catch (NotFoundException $e) {
+            // ensure if this really error with session
+            if ($this->causedBySessionNotFound($e)) {
+                if ($this->inTransaction()) {
+                    // if we inside transaction then throw abort exception
+                    throw new AbortedException(self::SESSION_NOT_FOUND_CONDITION);
```
taka-oyama, line 506: This removes all context information given by `$e`. I think `$e` should be passed to AbortedException.

src/Colopl/Spanner/Connection.php
```diff
@@ -443,4 +462,83 @@ protected function runQueryCallback($query, $bindings, Closure $callback)
...
+        }
+    }
+
+    /**
+     * Check if this is "session not found" error
+     *
+     * @param  Throwable  $e
+     * @return boolean
+     */
+    public function causedBySessionNotFound($e)
```
taka-oyama, line 538:
```suggestion
    public function causedBySessionNotFound(Throwable $e): bool
```

src/Colopl/Spanner/Connection.php
```diff
@@ -82,6 +84,21 @@ class Connection extends BaseConnection
...
+
+    /**
+     * Try to maintain and then clear session pool on 'session not found' error
+     */
+    const CLEAR_SESSION_POOL = 'CLEAR_SESSION_POOL';
+
+    /**
+     * Used to detect specific exception
+     */
+    const SESSION_NOT_FOUND_CONDITION = 'Session does not exist';
```
taka-oyama, line 100:
```suggestion
    public const SESSION_NOT_FOUND_CONDITION = 'Session does not exist';
```

src/Colopl/Spanner/Connection.php
```diff
@@ -82,6 +84,21 @@ class Connection extends BaseConnection
...
 
+    /**
+     * Try to maintain session pool on 'session not found' error
+     */
+    const MAINTAIN_SESSION_POOL = 'MAINTAIN_SESSION_POOL';
+
+    /**
+     * Try to maintain and then clear session pool on 'session not found' error
+     */
+    const CLEAR_SESSION_POOL = 'CLEAR_SESSION_POOL';
```
taka-oyama, line 95:
```suggestion
    public const CLEAR_SESSION_POOL = 'CLEAR_SESSION_POOL';
```

src/Colopl/Spanner/Connection.php
```diff
@@ -82,6 +84,21 @@ class Connection extends BaseConnection
      */
     protected $sessionPool;
 
+    /**
+     * Try to maintain session pool on 'session not found' error
+     */
+    const MAINTAIN_SESSION_POOL = 'MAINTAIN_SESSION_POOL';
```
taka-oyama, line 90:
```suggestion
    public const MAINTAIN_SESSION_POOL = 'MAINTAIN_SESSION_POOL';
```

taka-oyama:
Hi thanks for the pull request! I have some review feedback. Please check it out.

src/Colopl/Spanner/Concerns/ManagesSessionPool.php
oprudkyi: Hi @taka-oyama 
I got phpstan error there (probably because phpstan version not fixed in composer.json it installed the most recent version of phpstan - you probably will get the same errors if delete composer.lock and rerun tests/install) 
```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Colopl/Spanner/Concerns/ManagesSessionPool.php (in context of class Colopl\Spanner\Connection)                                                            
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------- 
  84     Parameter #1 $callback of method Illuminate\Support\Collection<(int|string),mixed>::map() expects callable(mixed, (int|string)): Colopl\Spanner\Session,  
         Closure(Google\Cloud\Spanner\V1\Session): Colopl\Spanner\Session given.                                                                                   
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------- 
```
and when I removed ProtobufSpannerSession then error changed to 
```
 ------ --------------------------------------------------------------------------------------------------------------------------------- 
  Line   Colopl/Spanner/Concerns/ManagesSessionPool.php (in context of class Colopl\Spanner\Connection)                                   
 ------ --------------------------------------------------------------------------------------------------------------------------------- 
  85     Parameter #1 $protobufSession of class Colopl\Spanner\Session constructor expects Google\Cloud\Spanner\V1\Session, mixed given.  
 ------ --------------------------------------------------------------------------------------------------------------------------------- 

```

and according to https://phpstan.org/writing-php-code/narrowing-types
the only suggested way - explicit checking of types - via assert or directly

src/Colopl/Spanner/Concerns/ManagesStaleReads.php
oprudkyi: to mute phpstan error
```
 ------ ------------------------------------------------------------------------------------------------------------- 
  Line   Colopl/Spanner/Concerns/ManagesStaleReads.php (in context of class Colopl\Spanner\Connection)                
 ------ ------------------------------------------------------------------------------------------------------------- 
  73     Method Colopl\Spanner\Connection::selectOneWithTimestampBound() should return array|null but returns mixed.  
 ------ ------------------------------------------------------------------------------------------------------------- 
```

src/Colopl/Spanner/Concerns/ManagesTransactions.php
oprudkyi: this is made specially to allow to run rollback without error
```
				} catch (Throwable $e) {
					$this->rollBack();
```
otherwise there is dangling transaction  and everything is breaking 
there is test for such use-case - testInTransactionSessionNotFoundHandledError

in more general sense - if session if lost, there are no ways to rollback transaction at all, the only manageable way - quietly abort current transaction and then rerun everything again

src/Colopl/Spanner/Connection.php
oprudkyi: I force-pushed commit with fix

src/Colopl/Spanner/Connection.php
oprudkyi: I force-pushed commit with fix

src/Colopl/Spanner/Connection.php
oprudkyi: I force-pushed commit with fix

src/Colopl/Spanner/Connection.php
oprudkyi: I force-pushed commit with fix

src/Colopl/Spanner/Connection.php
oprudkyi: fixed in force-pushed pr

src/Colopl/Spanner/Concerns/ManagesStaleReads.php
oprudkyi: fixed in force-pushed pr

src/Colopl/Spanner/Concerns/ManagesTransactions.php
```diff
@@ -44,29 +45,31 @@ trait ManagesTransactions
...
+				} catch (Throwable $e) {
+					$this->rollBack();
+					throw $e;
+				}
+			}, ['maxRetries' => $attempts - 1]);
+
+			$this->fireConnectionEvent('committed');
+
+			return $return;
+		});
```
taka-oyama, line 72: Please fix the tabs here to spaces as well.

src/Colopl/Spanner/Concerns/ManagesStaleReads.php
taka-oyama: Would adding `/** @var array<mixed>|null $result */` resolve the error as well?
I would like to avoid running asertions when we can get the same results without hindering performance.

src/Colopl/Spanner/Concerns/ManagesTransactions.php
taka-oyama: Thank you. That makes sense. Could you add that last sentence as a comment?

src/Colopl/Spanner/Connection.php
```diff
@@ -443,4 +462,83 @@ protected function runQueryCallback($query, $bindings, Closure $callback)
 
         return $result;
     }
+
+    /**
+     * Returns current mode
+     *
+     * @return string
+     */
+    protected function getSessionNotFoundMode()
```
taka-oyama, line 471:
```suggestion
    protected function getSessionNotFoundMode(): string
```

src/Colopl/Spanner/Concerns/ManagesTransactions.php
```diff
@@ -158,7 +161,14 @@ protected function performRollBack($toLevel)
 
         if ($this->currentTransaction !== null) {
             if ($this->currentTransaction->state() === Transaction::STATE_ACTIVE) {
-                $this->currentTransaction->rollBack();
+                try {
+                    $this->currentTransaction->rollBack();
+                } catch (NotFoundException $e) {
+                    // ignore session not found error
+                    if (empty($this->getSessionNotFoundMode()) || !$this->causedBySessionNotFound($e)) {
```
taka-oyama, line 168: I think it's safe to remove `empty($this->getSessionNotFoundMode())` since it will always contain a value.

README.md
```diff
@@ -236,6 +236,68 @@ In order to improve the performance of the first connection per request, we use
...
+[happen](https://cloud.google.com/spanner/docs/sessions#handle_deleted_sessions):
+
+ - Scripts that idle too long - for example, a [Laravel queue worker](https://laravel.com/docs/9.x/queues#the-queue-work-command)
+or anything that doesn't call Spanner frequently enough (more than once an hour).
+ - The session is more than 28 days old.
+ - Some random flukes on Google's side.
+
+The errors can be handled by one of the supported modes:
+
+- **MAINTAIN_SESSION_POOL** - When the 'session not found' error is encountered, the library tries to disconnect,
```
taka-oyama, line 251: I think the behavior for `CLEAR_SESSION_POOL` is good enough and this mode unnecessary.
By only having one way of doing things would make the whole thing simpler and easier to understand.
What do you think?

src/Colopl/Spanner/Concerns/ManagesSessionPool.php
taka-oyama: I'm starting to think phpstan level 9 might be too high for laravel based packages.
Would it bother you if I lowered it to 8? In which case I think this error will not appear here.

taka-oyama: I just found out that asserts can be tuned to not have performance impact on production so this code is OK.

https://www.php.net/manual/en/function.assert.php#function.assert.expectations

src/Colopl/Spanner/Concerns/ManagesStaleReads.php
oprudkyi: This is why I used assert(), it has special feature in recent php - any code inside it doesn't run in production mode
from https://www.php.net/manual/en/ini.core.php#ini.zend.assertions
>When set to -1, assertion code will not be generated, making the assertions zero-cost (production mode).
but I will do as you prefer

taka-oyama: Yeah. I just read up on this. Sorry for the confusion. This can stay as is.

src/Colopl/Spanner/Concerns/ManagesTransactions.php
oprudkyi: sure, also I found edge case when exception anyway should be called - i.e. on explicit rollback inside transaction({}), 
I am adding test-case and code for handling

src/Colopl/Spanner/Concerns/ManagesTransactions.php
oprudkyi: I thought to allow disabling mode by simple 
```
   	        'sessionNotFoundErrorMode' => false,
```
I am going to redo it to 
   	        'sessionNotFoundErrorMode' => 'THROW_EXCEPTION',

src/Colopl/Spanner/Concerns/ManagesTransactions.php
taka-oyama: Sorry for the ambiguity. The entire fix here is indented with tabs. Please replace the tabs with 4 spaces.

README.md
oprudkyi: Clearing session pool (while really is needed in very rare occasions)  is very 'hard' operation, which implicitly affects all processes that use session poll, i.e. for example if there 50 (httpd/php-fpm) workers that share session pool, then after clearing they all simultaneously will receive 'session not found' (because clearing deletes sessions) and end with restarting potentially heavy and long running scripts
so MAINTAIN_SESSION_POOL is the most safe mode - it still handles most of the cases (like idle workers), but do not introduce hard reset of sessions

so there are possible options
- left MAINTAIN_SESSION_POOL as safest way for users who doesn't want running into potentially risking hard reset
- left it and make it default 
- remove it, but provide users some level of control 
```
function setClearSessionPoolOnSessionNotFoundError(bool $value)
```
though it seems ugly and less clean

src/Colopl/Spanner/Connection.php
```diff
@@ -443,4 +462,83 @@ protected function runQueryCallback($query, $bindings, Closure $callback)
...
+
+    /**
+     * Handle "session not found" errors
+     *
+     * @template T
+     * @param  Closure(): T $callback
+     * @return T
+     * @throws InvalidArgumentException|NotFoundException|AbortedException
+     */
+    protected function sessionNotFoundWrapper(Closure $callback)
```
taka-oyama, line 484: Can you rename this to something like
```suggestion
    protected function withSessionHandling(Closure $callback): mixed
```

README.md
taka-oyama: I understood your concern. I now believe this can be as is.

However, now I'm thinking the opposite. Is `CLEAR_SESSION_POOL` necessary?

Wouldn't each session be retried eventually anyway?

oprudkyi: > Is CLEAR_SESSION_POOL necessary
yes, it is necessary at least in my use case - without it there is error once a week or two and maintain() didn't help

taka-oyama: So I'm guessing the error still occurred because in a case where `MAINTAIN_SESSION_POOL` is used, the handler only handles the error once, and upon reconnection, the logic probably choose another session that was also expired and errored out.
Is my understanding correct?

oprudkyi: no, There are rare cases when google spanner deletes a session entirely, while the session is still accounted in session pool as active, so maintain and reconnection just reload it again and again. and the only way to fix it is deletion of the cache   file manually (or via code, or restarting docker/pod)

src/Colopl/Spanner/Connection.php
oprudkyi: fixed in recent commit

README.md
oprudkyi: `MAINTAIN_SESSION_POOL` set as default mode

src/Colopl/Spanner/Concerns/ManagesTransactions.php
oprudkyi: done in recent commits

src/Colopl/Spanner/Concerns/ManagesTransactions.php
oprudkyi: done

src/Colopl/Spanner/Concerns/ManagesTransactions.php
oprudkyi: done, please check

src/Colopl/Spanner/Connection.php
```diff
@@ -443,4 +467,85 @@ protected function runQueryCallback($query, $bindings, Closure $callback)
...
+        }
+
+        try {
+            return $callback();
+        } catch (NotFoundException $e) {
+            // ensure if this really error with session
+            if ($this->causedBySessionNotFound($e)) {
+                if ($this->inTransaction()) {
+                    // if we inside transaction then throw abort exception
+                    throw new AbortedException(self::SESSION_NOT_FOUND_CONDITION, $e->getCode(), $e);
```
taka-oyama, line 513: I think we should throw this error after reconnecting.
That way we don't have to catch the AbortedException in the transaction code.

oprudkyi: Hi @taka-oyama 
this won't help - [Transaction](https://github.com/googleapis/google-cloud-php/blob/main/Spanner/src/Transaction.php) is tied to a session and stores it internally in private property. 
I tried to de-duplicate code in catch, please check 
https://github.com/colopl/laravel-spanner/pull/39/commits/df149c64df5c448f58c0e767b96465c2524c85a5

taka-oyama: So In that case, wouldn't it make sense to just throw the NotFoundException?

Wouldn't throwing AbortedException here [cause the transaction to retry](https://github.com/googleapis/google-cloud-php/blob/main/Spanner/src/Database.php#L864) with the same session?

oprudkyi: >So In that case, wouldn't it make sense to just throw the NotFoundException?
no, this will not work with reentrancy of withSessionNotFoundHandling

>Wouldn't throwing AbortedException here [cause the transaction to retry](https://github.com/googleapis/google-cloud-php/blob/main/Spanner/src/Database.php#L864) with the same session?

yes, I am going to reimplement with special exception class , something like TransactionRestartException

oprudkyi: Hi @taka-oyama 
I have implemented it without extra exception - just by handling ordinal QueryException

taka-oyama: Thank you for the fix!

taka-oyama:
Thank you for your contribution!

taka-oyama:
I will release the next version once I fix the Phpstan errors.

oprudkyi:
Hi @taka-oyama 

Thank you. 
could you also please create branch v3 from https://github.com/colopl/laravel-spanner/tree/v3.9.3
so it will be possible to backport to older laravel ?

taka-oyama:
Hi, thanks for the patience.

Yes, I will backport it.

I couldn't find time this week so I will do it on monday.

oprudkyi:
HI @taka-oyama 
I thought to make a backport by self - the my whole endeavor was to get universal solution for Laravel 8.
But any solution will do.

taka-oyama:
Hi. I released v4.1.0 and 3.10.0 (for laravel 8 or below), please check it out.

taka-oyama:
[Asking google to fix this upstream](https://github.com/googleapis/google-cloud-php/issues/5827), so we don't have to do this hack.

PR #41, opened 2022-06-15
fix: incorrect spec of partitionedDelete()

oprudkyi:
it seems typo, partitionedDelete() doesn't have any params
https://github.com/colopl/laravel-spanner/blob/ab3a6e13182037d8316ebf3ddefefde8bd82cf4e/src/Colopl/Spanner/Query/Concerns/UsesPartitionedDml.php#L45
```
    public function partitionedDelete()
    {
```


taka-oyama:
Hi, thank you for the fix!

PR #42, opened 2022-08-17
Feature: implement truncate

taka-oyama:


tests/Colopl/Spanner/Query/BuilderTest.php
```diff
@@ -819,4 +813,17 @@ public function testStaleReads()
...
         $this->assertEmpty($stalenessRow);
     }
+
+    public function testTruncate(): void
+    {
+        $conn = $this->getDefaultConnection();
+        $tableName = self::TABLE_NAME_USER;
+        $query = $conn->table($tableName);
+
+        $query->insert(['userId' => $this->generateUuid(), 'name' => 'first']);
```
halnique, line 823: It is better to assert that the inserted data exists before TRUNCATE, as in the DELETE test case. Because we don't know if the data was truncated as expected or if it was never in there in the first place.

```suggestion
        $insertData = ['userId' => $this->generateUuid(), 'name' => 'first'];
        $query->insert($insertData);
        $this->assertDatabaseHas(self::TABLE_NAME_USER, $insertData);
```

halnique (approved):
LGTM

PR #45, opened 2022-11-28
fix: session pool distinguishes connection

taka-oyama:
They were all managed in one file. They should be separated since they might be pointing to different instances.


taka-oyama:
The cache keys were unique to project/instance/database. This was not necessary.

PR #46, opened 2022-11-28
fix: session clear was not working correctly on CLI sessions

taka-oyama:
Addresses https://github.com/googleapis/google-cloud-php/issues/5567.

Unfortunately, Google is unresponsive so we have to take care of it on our end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# v5.0.0
+
+Changed
+- `Colopl\Spanner\Session` has been renamed to `Colopl\Spanner\SessionInfo`.
+- Default SessionPool was changed from `Google\Cloud\Spanner\Session\CacheSessionPool` to `Colopl\Spanner\CacheSessionPool`.
+
+Fixed
+- SessionPool was not cleared if php terminated immediately after calling `CacheSessionPool::clear`.
+
 # v4.2.0
 
 Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Changed
 - `Colopl\Spanner\Session` has been renamed to `Colopl\Spanner\SessionInfo`.
-- Default SessionPool was changed from `Google\Cloud\Spanner\Session\CacheSessionPool` to `Colopl\Spanner\CacheSessionPool`.
+- Default SessionPool was changed from `Google\Cloud\Spanner\Session\CacheSessionPool` to `Colopl\Spanner\CacheSessionPool` to address an [unresolved issue on google's end](https://github.com/googleapis/google-cloud-php/issues/5567).
 
 Fixed
 - SessionPool was not cleared if php terminated immediately after calling `CacheSessionPool::clear`.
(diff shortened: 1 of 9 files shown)


src/Colopl/Spanner/CacheSessionPool.php
```diff
@@ -0,0 +1,959 @@
...
+        $this->config = $config + self::$defaultConfig;
+        $this->validateConfig();
+    }
+
+    /**
+     * Acquire a session from the pool.
+     *
+     * @param string $context The type of session to fetch. May be either `r`
+     *        (READ) or `rw` (READ/WRITE). **Defaults to** `r`.
+     * @return Session
```
tyabu12, line 141:
```suggestion
     * @return SpannerSession
```

tyabu12:
Is there any reason why you didn't inherit from the `Google\Cloud\Spanner\Session\CacheSessionPool` and only overwrite the keepAlive method?

It is likely to leak when there is a change in the `GoogleCloudCloudSpanner\Session\CacheSessionPool` in the future.

taka-oyama:
Keepalive is not what we are overriding, its `deleteSessions`, which is a private method.

halnique (commented):
What about using [cweagans/composer-patches](https://github.com/cweagans/composer-patches) to apply patches instead of copying the file?

taka-oyama:
This requires that all projects that use it add the following lines to their composer.json. 

https://github.com/cweagans/composer-patches#allowing-patches-to-be-applied-from-dependencies

This is not a good solution since it requires users to change their composer.json manually.

taka-oyama:
Cleaned up commits so diff is minimal

taka-oyama:
> What about using [cweagans/composer-patches](https://github.com/cweagans/composer-patches) to apply patches instead of copying the file?

I just noticed you also need to `apk install patch` which is not installed by default on alpine.

I think just copying the original is fine here since google hasn't updated CacheSessionPool in ages.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# v5.0.0
+
+Changed
+- `Colopl\Spanner\Session` has been renamed to `Colopl\Spanner\SessionInfo`.
+- Default SessionPool was changed from `Google\Cloud\Spanner\Session\CacheSessionPool` to `Colopl\Spanner\CacheSessionPool`.
```
halnique, line 5: How about adding a description of googleapis/google-cloud-php#5567 and googleapis/google-cloud-php#5595 as related issues here?

halnique (approved):

CHANGELOG.md
taka-oyama: I'll add https://github.com/googleapis/google-cloud-php/issues/5567.

https://github.com/googleapis/google-cloud-php/issues/5595 is not handled here so that will not be included.

taka-oyama: done

halnique (approved):
LGTM

tyabu12 (approved):
LGTM

PR #49, opened 2022-12-01
feature: array types should be parsed in Grammar

taka-oyama:


taka-oyama:
Also fixes bug where stringArray had a different signature from string

halnique (approved):
LGTM

PR #50, opened 2022-12-01
chore: remove useless namespace folders

taka-oyama:


composer.json
```diff
@@ -25,12 +25,12 @@
...
   "autoload": {
     "psr-4": {
-      "Colopl\\Spanner\\": "src/Colopl/Spanner"
+      "Colopl\\Spanner\\": "src"
     }
   },
   "autoload-dev": {
     "psr-4": {
-      "Colopl\\Spanner\\Tests\\": "tests/Colopl/Spanner"
+      "Colopl\\Spanner\\Tests\\": "tests/src"
```
halnique, line 33: nits

https://www.php-fig.org/psr/psr-4/examples/#class-example

There seems to be no clear rule, but I believe that in many cases the src directory and depth are aligned.
How about moving the test code directory up one level?

taka-oyama: Done.

> many cases the src directory and depth are aligned.

I'm not sure about this.
I think everyone just choose whatever fits their needs.
These commonly use packages all do their own thing.

- https://github.com/phpstan/phpstan-src
- https://github.com/symfony/http-kernel
- https://github.com/sebastianbergmann/phpunit

halnique (approved):
LGTM

PR #56, opened 2022-12-08
Feature/add spanner:sessions command

taka-oyama:
This command does not work when used with emulator


src/Console/SessionsCommand.php
```diff
@@ -0,0 +1,47 @@
...
+                $data = [];
+                $this->info("{$connection->getName()} contains {$count} session(s).");
+                foreach ($sessions as $session) {
+                    $data[] = [
+                        $session->getName(),
+                        $session->getCreatedAt(),
+                        $session->getLastUsedAt(),
+                    ];
+                }
+                $this->table($headers, $data);
```
halnique, line 43: nits
If `$data` is empty, only the header is output, which is a bit uncool.

halnique (approved):
LGTM

PR #57, opened 2022-12-12
feature: add refresh option for warmup

taka-oyama:
It will allow warmup to create fresh sessions instead of reusing existing ones.


halnique (approved):
LGTM

PR #59, opened 2022-12-15
Feat: add action classes for interleave and index

taka-oyama:
For autocompletion in PhpStorm

### DIFF
--- a/src/Schema/IndexDefinition.php
+++ b/src/Schema/IndexDefinition.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Colopl\Spanner\Schema;
+
+use Illuminate\Support\Fluent;
+
+/**
+ * @method $this interleave(string $table)
+ * @method $this nullFiltered()
+ * @method $this storing(string[] $columns)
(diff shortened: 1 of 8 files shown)


src/Schema/IndexDefinition.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Colopl\Spanner\Schema;
+
+use Illuminate\Support\Fluent;
+
+/**
+ * @method $this interleave(string $table)
+ * @method $this nullFiltered()
+ * @method $this storing(string[] $columns)
```
tyabu12, line 10: Unrelated to this PR, but It might be nice to explain the nullFiltered and storing index in README.md

tyabu12 (approved):
LGTM

src/Schema/IndexDefinition.php
taka-oyama: docs have been added.

tyabu12: Perfect, thank you.

PR #63, opened 2022-12-28
feat: use symfony's FilesystemAdapter for auth and session pool caching.

taka-oyama:
Changes
- Auth and session pool no longer use the custom FileCacheAdapter and uses Symfony's FilesystemAdapter instead.
- Path for auth and session pool files have moved from `storage/framework/cache/spanner` to `storage/framework/spanner/{auth|session}`.


src/SpannerServiceProvider.php
```diff
@@ -71,22 +77,22 @@ protected function parseConfig(array $config, $name)
...
-     * @param array $sessionPoolConfig
+     * @param array<string, mixed> $sessionPoolConfig
      * @return SessionPoolInterface
      */
     protected function createSessionPool(array $sessionPoolConfig): SessionPoolInterface
     {
-        $cachePath = storage_path(implode(DIRECTORY_SEPARATOR, ['framework', 'cache', 'spanner']));
-        return new CacheSessionPool(new FileCacheAdapter('session', $cachePath), $sessionPoolConfig);
+        $cachePath = storage_path(implode(DIRECTORY_SEPARATOR, ['framework', 'spanner']));
+        $adapter = new FilesystemAdapter('session', 0, $cachePath);
```
halnique, line 86:
```suggestion
        $adapter = new FilesystemAdapter(namespace: 'session', directory: $cachePath);
```

How about using `Named Arguments`?

src/SpannerServiceProvider.php
```diff
@@ -71,22 +77,22 @@ protected function parseConfig(array $config, $name)
...
 
     /**
      * @return CacheItemPoolInterface
      */
     protected function createAuthCache(): CacheItemPoolInterface
     {
-        $cachePath = storage_path(implode(DIRECTORY_SEPARATOR, ['framework', 'cache', 'spanner']));
-        return new FileCacheAdapter('auth', $cachePath);
+        $cachePath = storage_path(implode(DIRECTORY_SEPARATOR, ['framework', 'spanner']));
+        return new FilesystemAdapter('auth', 0, $cachePath);
```
halnique, line 96:
```suggestion
        return new FilesystemAdapter(namespace: 'auth', directory: $cachePath);
```

src/SpannerServiceProvider.php
taka-oyama: I don't understand the need for it?

halnique: The `$defaultLifetime` (the second argument) is the same as the default value, but the intent is to explicitly leave it to the default value since it is not of interest in this use case.

taka-oyama: Actually, unlimited lifetime is what is expected here, so explicitly defining it would be a good thing here though?

halnique: Okay, if it is an intentional designation, no problem.
I thought I was forced to specify the default value of `$defaultLifetime` in order to specify `$directory` (the third argument), but I was mistaken.

taka-oyama: Apologies for the confusion.

Before this fix, the FileCacheAdapter we used did not allow you to specify the lifetime and was expected to be unlimited (although internally, it did not specify so it was using the default behavior).

In this case, either solution works, because the default fall back is unlimited anyway, but I figured explicit would be best here.

halnique (approved):
LGTM

tyabu12 (approved):
LGTM

PR #65, opened 2023-01-06
feat: support Schema\Builder::getAllTables()

taka-oyama:


src/Query/Processor.php
```diff
@@ -42,6 +42,18 @@ public function processSelect(Builder $query, $results): array
...
     }
 
+    /**
+     * @param list<array{ TABLE_NAME: string }> $results
+     * @return list<string>
+     */
+    public function processGetAllTables(array $results): array
+    {
+        return array_map(
+            static fn(array $result) => $result['TABLE_NAME'],
```
tyabu12, line 52:
```suggestion
            static fn(array $result): string => $result['TABLE_NAME'],
```

src/Schema/Grammar.php
```diff
@@ -48,6 +48,14 @@ public function compileTableExists()
         return 'select * from information_schema.tables where table_schema = \'\' and table_name = ?';
     }
 
+    /**
+     * @return string
+     */
+    public function compileGetAllTables()
+    {
+        return 'select * from information_schema.tables where table_schema = \'\'';
```
tyabu12, line 56: This SQL includes view table.  

I think it should be to add an additional table_type condition to exclude it as well as [MySqlGrammer::compileGetAllTables()](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php#L60-L68).

https://cloud.google.com/spanner/docs/information-schema#tables

```suggestion
        return 'select * from information_schema.tables where table_schema = \'\' and table_type = \'BASE TABLE\'';
```

tyabu12 (changes requested):

src/Query/Processor.php
taka-oyama: I usually don't add return types since phpstan should be taking care of type checking.

tyabu12 (changes requested):
Would you add about `Schema\Builder::getAllTables()` support to CHANGELOG.md?

tyabu12 (approved):
LGTM

PR #66, opened 2023-01-11
fix: session clear was not working correctly on CLI sessions

taka-oyama:
A better implementation of https://github.com/colopl/laravel-spanner/pull/46.

Has no beaking change so it can be applied to v4.

Implementation is a direct copy of https://github.com/googleapis/google-cloud-php/pull/5774.

The only exception are the lines added to `Connection`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 Added
 - Support Schema\Builder::getAllTables()
 
+Changed
+- Default SessionPool was changed from `Google\Cloud\Spanner\Session\CacheSessionPool` to `Colopl\Spanner\Session\CacheSessionPool` to patch an [unresolved issue on google's end](https://github.com/
+
+Fixed
+- SessionPool was not cleared if php terminated immediately after calling `CacheSessionPool::clear`.
+
 # v4.3.0
 
 Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Added
 - Support Schema\Builder::getAllTables()
 
 Changed
-- Default SessionPool was changed from `Google\Cloud\Spanner\Session\CacheSessionPool` to `Colopl\Spanner\Session\CacheSessionPool` to patch an [unresolved issue on google's end](https://github.com/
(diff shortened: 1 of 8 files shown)


src/Session/DeleteOperation.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->promises = $promises;
+    }
+
+    /**
+     * @return void
+     */
+    public function waitUntilComplete()
+    {
+        foreach ($this->promises as $promise) {
+            $promise->wait(false);
```
taka-oyama, line 43: Disable unwrapping of rejected promises so it doesn't stop the code.

phpstan.neon
```diff
@@ -27,3 +27,5 @@ parameters:
           path: src/Concerns/ManagesTransactions.php
         - message: '#^Parameter \#1 \$connection of method Illuminate\\Database\\DatabaseTransactionsManager::.+\(\) expects string, string\|null given\.$#'
           path: src/Concerns/ManagesTransactions.php
+        - message: "#.*#"
```
tyabu12, line 30: How about using excludePaths.analyse ?

https://phpstan.org/user-guide/ignoring-errors#excluding-whole-files

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 Added
 - Support Schema\Builder::getAllTables()
 
+Changed
+- Default SessionPool was changed from `Google\Cloud\Spanner\Session\CacheSessionPool` to `Colopl\Spanner\Session\CacheSessionPool` to patch an [unresolved issue on google's end](https://github.com/
```
tyabu12, line 7: it seems that the issue link is broken.

tyabu12 (changes requested):

src/Session/DeleteOperation.php
```diff
@@ -0,0 +1,46 @@
+<?php
+/**
+ * Copyright 2016 Google Inc.
```
halnique, line 3: Is this correct?

halnique (commented):
Does it not apply to the master branch?

phpstan.neon
taka-oyama: Looks good, they have been added.

src/Session/DeleteOperation.php
taka-oyama: fixed.

halnique (approved):
LGTM

It would be a good idea to change the master branch (version 5.0 or later) to this approach as well.
Will this be changed in another PR?

taka-oyama:
Yes, that's the plan. I think I will just remove the old fix from v5 branch.

PR #67, opened 2023-01-12
feature: add sort and order option to spanner:sessions

taka-oyama:


src/Console/SessionsCommand.php
```diff
@@ -44,21 +51,59 @@ public function handle(DatabaseManager $db): void
...
+    protected function makeSessionData(Connection $connection): Collection
+    {
+        $descending = $this->getOrder() === 'desc';
+
+        return $connection->listSessions()
+            ->sortBy(fn(Session $s) => $this->getSortValue($s), descending: $descending)
+            ->map(static fn(Session $s) => [
+                $s->getName(),
+                $s->getCreatedAt()->format('Y-m-d H:i:s'),
+                $s->getLastUsedAt()->format('Y-m-d H:i:s'),
```
tyabu12, line 76: How about using associative array?
`handle()` function no longer needs to know the `$header` and the order of the results.

```suggestion
                'name' => $s->getName(),
                'createdAt' => $s->getCreatedAt()->format('Y-m-d H:i:s'),
                'lastUsedAt' => $s->getLastUsedAt()->format('Y-m-d H:i:s'),
```

tests/Console/SessionsCommandTest.php
```diff
@@ -18,9 +18,28 @@
...
+use Google\Cloud\Spanner\Session\SessionPoolInterface;
 
 class SessionsCommandTest extends TestCase
 {
+    /**
+     * @param Connection $connection
+     * @param int $amount
+     * @return list<Session>
+     */
+    protected function createSessions(Connection $connection, int $amount)
```
tyabu12, line 33:
```suggestion
    protected function createSessions(Connection $connection, int $amount): array
```

tyabu12 (changes requested):

src/Console/SessionsCommand.php
taka-oyama: sounds good.

taka-oyama: Implemented.

src/Console/SessionsCommand.php
```diff
@@ -39,26 +46,63 @@ public function handle(DatabaseManager $db): void
...
+     * @return string
+     */
+    protected function getOrder(): string
+    {
+        $order = $this->option('order') ?? 'desc';
+        assert(is_string($order));
+
+        $order = strtolower($order);
+
+        if (!in_array($order, ['asc', 'desc'])) {
```
tyabu12, line 101:
```suggestion
        if (!in_array($order, ['asc', 'desc'], true)) {
```

src/Console/SessionsCommand.php
```diff
@@ -17,13 +17,20 @@
...
+use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
+use RuntimeException;
 
 class SessionsCommand extends Command
 {
-    protected $signature = 'spanner:sessions {connections?* : The database connections to query}';
+    protected $signature = 'spanner:sessions {connections?* : The database connections to query}
+               {--sort= : Name of column to be sorted (default: LastUsedAt)}
+               {--order= : sort order as "ASC" or "DESC" (default: DESC)}';
```
tyabu12, line 33: >You may assign default values to options by specifying the default value after the option name.
https://laravel.com/docs/9.x/artisan#options-with-values

```suggestion
               {--sort=LastUsedAt : Name of column to be sorted}
               {--order=DESC : sort order as "ASC" or "DESC"}';
```

tyabu12 (changes requested):

src/Console/SessionsCommand.php
taka-oyama: fixed.

src/Console/SessionsCommand.php
taka-oyama: fixed.

src/Console/SessionsCommand.php
```diff
@@ -39,26 +46,63 @@ public function handle(DatabaseManager $db): void
...
+    /**
+     * @param Session $session
+     * @return string
+     */
+    protected function getSortValue(Session $session): string
+    {
+        $sort = $this->option('sort');
+        assert(is_string($sort));
+        $method = 'get' . Str::studly($sort);
+        return (string) $session->$method();
```
halnique, line 88: How about checking with `method_exists()` before calling `$method`?

If possible, it would be easier to use if the value that can be specified as an option could be included in the `$description` or exception message.

src/Console/SessionsCommand.php
```diff
@@ -17,13 +17,20 @@
...
+use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
+use RuntimeException;
 
 class SessionsCommand extends Command
 {
-    protected $signature = 'spanner:sessions {connections?* : The database connections to query}';
+    protected $signature = 'spanner:sessions {connections?* : The database connections to query}
+               {--sort=LastUsedAt : Name of column to be sorted}
+               {--order=desc : sort order as "asc" or "desc"}';
```
tyabu12, line 33:
```suggestion
               {--order=desc : Sort order as "asc" or "desc"}';
```

src/Console/SessionsCommand.php
taka-oyama: Added columns to description and also fail on mismatch.

src/Console/SessionsCommand.php
```diff
@@ -39,26 +46,67 @@ public function handle(DatabaseManager $db): void
...
+    /**
+     * @param Session $session
+     * @return string
+     */
+    protected function getSortValue(Session $session): string
+    {
+        $sort = $this->option('sort');
+        assert(is_string($sort));
+        return match (Str::camel($sort)) {
+            'name' => $session->getName(),
```
halnique, line 88:
```suggestion
            'Name' => $session->getName(),
```

halnique (changes requested):

src/Console/SessionsCommand.php
```diff
@@ -39,26 +46,67 @@ public function handle(DatabaseManager $db): void
...
+
+    /**
+     * @param Session $session
+     * @return string
+     */
+    protected function getSortValue(Session $session): string
+    {
+        $sort = $this->option('sort');
+        assert(is_string($sort));
+        return match (Str::camel($sort)) {
```
tyabu12, line 87: I think that this should be not lower camel case but upper camel case.

```suggestion
        return match (Str:: studly($sort)) {
```

tyabu12 (changes requested):

src/Console/SessionsCommand.php
```diff
@@ -39,26 +46,67 @@ public function handle(DatabaseManager $db): void
...
+     * @return string
+     */
+    protected function getSortValue(Session $session): string
+    {
+        $sort = $this->option('sort');
+        assert(is_string($sort));
+        return match (Str::studly($sort)) {
+            'Name' => $session->getName(),
+            'Created' => (string) $session->getCreatedAt(),
+            'LastUsed' => (string) $session->getLastUsedAt(),
```
halnique, line 90: Is there a reason you are using `format('Y-m-d H:i:s')` and `(string)` casts?
'Y-m-d H:i:s' is the default format of `Carbon`, so I would expect the result to be the same.

tests/Console/SessionsCommandTest.php
```diff
@@ -18,9 +18,28 @@
...
 class SessionsCommandTest extends TestCase
 {
+    /**
+     * @param Connection $connection
+     * @param int $amount
+     * @return list<Session>
+     */
+    protected function createSessions(Connection $connection, int $amount): array
+    {
+        $pool = $connection->getSpannerDatabase()->sessionPool() ?? throw new Exception('unreachable');
```
halnique, line 35: I think unchecked exception (or `RuntimeException` in this case) would be more appropriate for `unreachable' exceptions.

halnique (commented):
I have some `nits` comments.

src/Console/SessionsCommand.php
taka-oyama: I have changed it to cast instead.

tests/Console/SessionsCommandTest.php
taka-oyama: I figured, this would suffice since it's a simple check to prevent phpstan from raising an erroring here.

In an optimal scenario, I would create a `UnreachableException` and use that, but that would be overkill for a simple check within a test. 

I have changed it to `RuntimeException ` per request.

tyabu12 (approved):
LGTM

PR #68, opened 2023-01-17
feature: change default sessionNotFoundErrorMode to CLEAR_SESSION_POOL

taka-oyama:
I wasn't fully confident at the time, but I think it should be safe to assume it's working now.


halnique (approved):
LGTM

PR #69, opened 2023-01-27
fix: Transaction was not cleared on NotFoundException

taka-oyama:
See test for when this might occur.

### DIFF
--- a/tests/ConnectionTest.php
+++ b/tests/ConnectionTest.php
@@ -355,7 +355,10 @@ class ConnectionTest extends TestCase
         $conn = $this->getDefaultConnection();
 
         try {
-            $conn->transaction(function () {
+            $conn->transaction(function (Connection $conn) {
+                self::assertTrue($conn->inTransaction());
+                self::assertNotNull($conn->getCurrentTransaction());
+                self::assertSame(1, $conn->transactionLevel());
                 throw new NotFoundException('NG');
             });
         } catch(NotFoundException) {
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -9,4 +9,4 @@ services:
     depends_on:
       - emulator
   emulator:
-    image: gcr.io/cloud-spanner-emulator/emulator
+    image: gcr.io/cloud-spanner-emulator/emulator:1.4.9
```
taka-oyama, line 12: Fix spanner-emulator version to 1.4.9 until issue is fixed. 
https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/98

tests/ConnectionTest.php
```diff
@@ -347,4 +349,21 @@ public function testEventListenOrder(): void
         $this->assertEquals(QueryExecuted::class, $receivedEventClasses[1]);
         $this->assertEquals(TransactionCommitted::class, $receivedEventClasses[2]);
     }
+
+    public function test_Connection_transaction_reset_on_exceptions(): void
+    {
+        $conn = $this->getDefaultConnection();
+
+        try {
+            $conn->transaction(function () {
```
halnique, line 358:
```suggestion
            $conn->transaction(function (Connection $conn) {
                self::assertTrue($conn->inTransaction());
                self::assertNotNull($conn->getCurrentTransaction());
                self::assertSame(1, $conn->transactionLevel());
```
How about asserting before throwing an exception to make sure that the transaction is definitely initiated?

zeriyoshi (approved):

halnique (approved):

PR #73, opened 2023-02-10
Breaking Change: Schema\Builder::getAllTables was returning wrong type

taka-oyama:
was list of string but should be row with name and type (differs for various databases)

### DIFF
--- a/src/Schema/Builder.php
+++ b/src/Schema/Builder.php
@@ -35,19 +35,13 @@ class Builder extends BaseBuilder
 
 
     /**
-     * @return list<string>
+     * @return list<array{ name: string, type: string }>
      */
     public function getAllTables()
     {
-        /** @var list<array{ TABLE_NAME: string }> $results */
-        $results = $this->connection->select(
+        return $this->connection->select(
             $this->grammar->compileGetAllTables()
         );
-
-        /** @var Processor $processor */
-        $processor = $this->connection->getPostProcessor();
-
-        return $processor->processGetAllTables($results);
     }
 
     /**
--- a/tests/Schema/BuilderTest.php
+++ b/tests/Schema/BuilderTest.php
@@ -197,11 +197,11 @@ class BuilderTest extends TestCase
             $table->primary('id');
         });
 
-        $tables = array_map(
-            static fn (array $row): string => $row['name'],
-            $sb->getAllTables(),
-        );
+        $tables = $sb->getAllTables();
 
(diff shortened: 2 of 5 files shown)


src/Schema/Builder.php
```diff
@@ -35,19 +35,13 @@ class Builder extends BaseBuilder
 
 
     /**
-     * @return list<string>
+     * @return list<array{ name: string, type: string }>
```
tyabu12, line 38: This is a breaking change. Could you add this to the CHANGELOG?

tests/Schema/BuilderTest.php
```diff
@@ -197,6 +197,11 @@ public function test_getAllTables(): void
             $table->primary('id');
         });
 
-        $this->assertContains(self::TABLE_NAME_CREATED, $sb->getAllTables());
+        $tables = array_map(
+            static fn (array $row): string => $row['name'],
```
tyabu12, line 201: How about adding an assertion about `$row['type']`?

tyabu12 (changes requested):

src/Schema/Builder.php
taka-oyama: added

tests/Schema/BuilderTest.php
taka-oyama: They have been added.

tyabu12 (approved):
LGTM

src/Schema/Builder.php
```diff
@@ -35,19 +35,13 @@ class Builder extends BaseBuilder
...
-        /** @var list<array{ TABLE_NAME: string }> $results */
-        $results = $this->connection->select(
+        return $this->connection->select(
             $this->grammar->compileGetAllTables()
         );
-
-        /** @var Processor $processor */
-        $processor = $this->connection->getPostProcessor();
-
-        return $processor->processGetAllTables($results);
```
halnique, line 50: https://github.com/colopl/laravel-spanner/blob/0a9ca9811486dfd45e5eabb2a5c72edd37c2e704/src/Query/Processor.php

Would you also remove `Processor::processGetAllTables`?
(It's a breaking change anyway...)

taka-oyama: fixed

zeriyoshi (approved):
LGTM

tyabu12 (approved):
LGTM

PR #74, opened 2023-02-15
fix: match forceIndex arg with base class (added in laravel 9.52)

taka-oyama:
This package broke on laravel 9.52.0 since it introduced their own version of forceIndex, which has a different arg signature.


taka-oyama:
Thank you. CHANGELOG has been added.

halnique (approved):

tyabu12 (approved):
LGTM

PR #78, opened 2023-02-21
fix: Expression should go through Grammar::getValue

taka-oyama:
No change in behavior. Makes it easier to transition to laravel 10.


tyabu12 (approved):
LGTM

PR #87, opened 2023-04-06
fix: `Connection::createDatabase` was not logging and was not triggering events for given statements

taka-oyama:


tests/Concerns/ManagesDataDefinitionsTest.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    public function test_createDatabase_with_statements(): void
+    {
+        $events = Event::fake([QueryExecuted::class]);
+
+        $conn = new Connection('test-instance', 'test_' . time(), '', ['instance' => 'test-instance']);
+
+        if (!empty(getenv('SPANNER_EMULATOR_HOST'))) {
+            $this->setUpEmulatorInstance($conn);
+        }
```
halnique, line 35: `setUpDatabaseOnce()` cannot be used?

```suggestion
        $this->setUpDatabaseOnce($conn);
```

taka-oyama: setUpDatabaseOnce will create a database, which is not what we want.

We want to create the instance but not the database.

halnique (approved):

PR #88, opened 2023-04-13
fix: emulator support for Connection::listSessions

taka-oyama:


halnique (approved):

halnique:
I apologize, I overlooked this earlier. It seems like it would be better to enable the following as well.

```php
    public function test_no_args(): void
    {
        if (getenv('SPANNER_EMULATOR_HOST')) {
            $this->markTestSkipped('Cannot list sessions on emulator');
        }
```

Shall we address this together with #89 ?

taka-oyama:
I have added the fix for it in #89.

PR #89, opened 2023-04-13
feature: separate connection per session so it is not shared

taka-oyama:


src/SpannerServiceProvider.php
```diff
@@ -87,22 +87,24 @@ protected function parseConfig(array $config, string $name): array
...
         $adapter = new FilesystemAdapter('session', 0, $cachePath);
         return new CacheSessionPool($adapter, $sessionPoolConfig);
     }
 
     /**
+     * @param string $name
      * @return CacheItemPoolInterface
      */
-    protected function createAuthCache(): CacheItemPoolInterface
+    protected function createAuthCache(string $name, ): CacheItemPoolInterface
```
tyabu12, line 105:
```suggestion
    protected function createAuthCache(string $name): CacheItemPoolInterface
```

tyabu12 (approved):
LGTM

halnique (approved):

PR #90, opened 2023-04-13
feature: use official CacheSessionPool from google/cloud-spanner

taka-oyama:


src/Concerns/ManagesSessionPool.php
```diff
@@ -42,11 +42,7 @@ trait ManagesSessionPool
     public function clearSessionPool(): void
     {
         $sessionPool = $this->getSpannerDatabase()->sessionPool();
-        $operation = $sessionPool?->clear();
-
-        if ($operation instanceof DeleteOperation) {
```
tyabu12, line 47: Cloud you remove use `Colopl\Spanner\Session\DeleteOperation` ?

src/Session/CacheSessionPool.php
```diff
@@ -1,1048 +0,0 @@
-<?php
```
tyabu12, line 1: Cloud you also remove excludePaths setting in phpstan.neon?

https://github.com/colopl/laravel-spanner/blob/master/phpstan.neon#L6-L8

src/Concerns/ManagesSessionPool.php
taka-oyama: done

src/Session/CacheSessionPool.php
taka-oyama: It has been removed. Thanks.

tyabu12 (approved):
LGTM

Issue #94, opened 2023-04-27
Incorrect dependencies


zeriyoshi:
v5.0.0 targets Laravel 10 or later, but requires PHP version 8.0 or later. This is incorrect; 8.1 or later is required.

Also, the PHP version specified in composer.json is too broad and may cause future incompatibilities, so you should only change to a minor version that is known to work.

PR #95, opened 2023-04-27
Fix incorrect and weak requirements

zeriyoshi:
Fix: #94 

P.S. Due to the nature of being hosted on Packagist, it is necessary to cancel the v5.0.0 release. Otherwise, it will be installed improperly in the event of future incompatibility with minor PHP upgrades.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     {"name": "Takayasu Oyama", "email": "t-oyama@colopl.co.jp"}
   ],
   "require": {
-    "php": "~8.1.0 || ~8.2.0",
+    "php": "^8.1",
     "ext-grpc": "*",
     "ext-json": "*",
     "laravel/framework": "~10.0",
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     {"name": "Takayasu Oyama", "email": "t-oyama@colopl.co.jp"}
   ],
   "require": {
-    "php": ">=8",
+    "php": "~8.1.0 || ~8.2.0",
     "ext-grpc": "*",
     "ext-json": "*",
     "laravel/framework": "~10.0",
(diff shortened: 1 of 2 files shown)


taka-oyama:
I'm not convinced by this change, I don't see this done in any of the major libraries.

- https://github.com/laravel/framework/blob/10.x/composer.json#L18
- https://github.com/symfony/symfony/blob/6.3/composer.json#L36
- https://github.com/phpstan/phpstan-src/blob/1.11.x/composer.json#L8
- https://github.com/rectorphp/rector/blob/main/composer.json#L10

So you have to be more specific about "future incompatibilities".

> it is necessary to cancel the v5.0.0 release

Nope, that's a major breakage and will break anyone currently using it.
You're talking about breaking existing users because there is a "potential for breakage".

If you have a problem with the current config, all that needs to be done is to release a patch version.

composer.json
```diff
@@ -8,7 +8,7 @@
     {"name": "Takayasu Oyama", "email": "t-oyama@colopl.co.jp"}
   ],
   "require": {
-    "php": ">=8",
+    "php": "~8.1.0 || ~8.2.0",
```
taka-oyama, line 11:
```suggestion
    "php": "^8.1",
```

taka-oyama (changes requested):

zeriyoshi:
@taka-oyama 
I understand. I think it's a good idea to match the standard of other projects for the time being.
It is certainly not a good idea to cancel a release and destroy the composer.lock of a project that is already using it.

On the other hand, it should be noted that PHP has been aggressively adopting disruptive changes in recent years and is not SemVer compliant. I think you should consider moving to a strict version specification when you support Laravel 11. If you agree, I will make it an issue for the record.

PR #98, opened 2023-05-01
feature: deprecate runDdl and runDdls

taka-oyama:
- [x] CHANGELOG


src/Concerns/ManagesDataDefinitions.php
```diff
@@ -19,27 +19,32 @@
...
 trait ManagesDataDefinitions
 {
     /**
+     * @deprecated use runDdlBatch() instead
      * @param string $ddl
      * @return LongRunningOperation
      */
     public function runDdl(string $ddl): LongRunningOperation
     {
+        trigger_deprecation('colopl/laravel-spanner', '5.1.1', 'runDdl() is deprecated. Use runDdlBatch() instead.');
```
tyabu12, line 36: According to CHANGELOG, this change looks like it will be included in v5.2.0.
Which version is the correct one?

taka-oyama: 5.2 is the correct version. this has been fixed. Thanks.

tyabu12 (approved):
LGTM

PR #99, opened 2023-05-01
feature: remove Connection's waitForOperation and isDoneOperation

taka-oyama:
Connection::isDoneOperation was not used anywhere. Connection::waitForOperation is only used in trait ManagesDataDefinitions, so it's been moved there as a protected method.

The trait ManagesLongRunningOperations is also removed.

- [x] CHANGELOG


tyabu12 (approved):
LGTM

PR #100, opened 2023-05-01
chore: refactoring PHPStan usage

zeriyoshi:
## Checklist

- ~[] CHANGELOG~

## Reference

In PHPStan 1.10.2 and later, a faster [`phpstan-baseline.php` format baseline file is available](https://phpstan.org/user-guide/baseline#php-baseline-format-instead-of-neon).

Change the baseline to an auto-generated one and remove the suppression of warnings about array datatype information to further clean up the code in the future.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -2,7 +2,7 @@ services:
   test:
     build:
       context: .
-    command: /bin/sh -c "composer install --no-progress && vendor/bin/phpstan --no-progress --memory-limit=-1 && vendor/bin/phpunit"
+    command: /bin/sh -c "composer install --no-progress && composer exec -- phpstan --no-progress --memory-limit=-1 && composer exec phpunit"
     volumes:
       - .:/project
     depends_on:
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -1,0 +1,670 @@
+<?php declare(strict_types = 1);
+
+$ignoreErrors = [];
+$ignoreErrors[] = [
+	'message' => '#^Method Colopl\\\\Spanner\\\\Connection\\:\\:__construct\\(\\) has parameter \\$config with no value type specified in iterable type array\\.$#',
+	'count' => 1,
+	'path' => __DIR__ . '/src/Connection.php',
+];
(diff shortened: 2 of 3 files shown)


taka-oyama:
The php version is only effective for large baseline. It's useless for the scale of this library.

I actually measured it. It's virtually the same.

php version
```
real	0m 35.47s
user	0m 31.23s
sys	0m 2.50s
```
neon version
```
real	0m 34.70s
user	0m 30.25s
sys	0m 2.74s
```

> further clean up the code in the future.

I don't plan on doing this anytime soon.

zeriyoshi:
@taka-oyama 
It should be noted that the current ignoreErrors does not specify the number of occurrences, and no warning is given when a new occurrence of the same error occurs. 

This will affect the quality of code written in the future and should be properly migrated to the auto-generated baseline.

Also, there is no advantage in using neon files when doing so.

If you still think this is unnecessary, please close the this issue.

taka-oyama:
If this contained the fixes for the warnings, I would have loved to merge it, but this PR just forces ME to do the fixing in the future.

I have priorities, and fixing small potential issues is very low on the list.

zeriyoshi:
> If this contained the actual fixes for the warnings, I would have loved to merge it, but this PR just forces ME to do the fixing in the future.

No, it is not. Only new implementations are mandated to be properly implemented. Is there any reason why you would not want to properly implement a new implementation?

If you want to put off dealing with it, you can easily regenerate it with `phpstan --generate-baseline=phpstan-baseline.php`. This is a total hassle-free solution rather than ignoring future defects.

taka-oyama:
That's not how this PR was initially explained.

Also, please stop doing multiple things at one in one single PR, it's really hard to read/understand.

zeriyoshi:
This PR only changes the way PHPStan is executed and should not be too difficult to read. There is no need to read the generated baseline.

As I have said from the beginning, this PR is intended to improve the quality of the code in the future, and changing the way PHPStan is used, i.e., how the baselines are generated, is a necessary step. How do you subdivide the changes any further?

Issue #113, opened 2023-06-12
Add support for Federation M2


taka-oyama:
See here for PR.

https://github.com/googleapis/google-cloud-php/pull/6300/files

PR #122, opened 2023-07-19
feat: add Connection methods selectWithOptions and cursorWithOptions

taka-oyama:
Must be implemented before implementing https://github.com/colopl/laravel-spanner/issues/113

- [x] CHANGELOG

### DIFF
--- a/src/Concerns/ManagesStaleReads.php
+++ b/src/Concerns/ManagesStaleReads.php
@@ -30,7 +30,7 @@ trait ManagesStaleReads
      * @param string $query
      * @param array<array-key, mixed> $bindings
      * @param TimestampBoundInterface|null $timestampBound
-     * @return Generator<int, list<mixed>|null>
+     * @return Generator<int, array<array-key, mixed>>
      */
     public function cursorWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): Generator
     {
--- a/src/Concerns/ManagesStaleReads.php
+++ b/src/Concerns/ManagesStaleReads.php
@@ -58,7 +58,7 @@ trait ManagesStaleReads
      */
     public function selectOneWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): ?array
     {
-        return $this->cursorWithTimestampBound($query, $bindings, $timestampBound)->current();
+        return $this->selectWithTimestampBound($query, $bindings, $timestampBound)[0] ?? null;
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/Connection.php
```diff
@@ -256,32 +256,44 @@ public function query(): QueryBuilder
...
+     * @param array<string, mixed> $options
+     * @return array<int, array<array-key, mixed>>
+     */
+    public function selectWithOptions(string $query, array $bindings, array $options): array
+    {
+        return $this->run($query, $bindings, function ($query, $bindings) use ($options): array {
+            return !$this->pretending()
+                ? iterator_to_array($this->executeQuery($query, $bindings, $options))
+                : [];
+        });
```
halnique, line 282:
```suggestion
        return iterator_to_array($this->cursorWithOptions($query, $bindings, $options));
```
How about using `cursorWithOptions`?

src/Query/Builder.php
```diff
@@ -132,12 +132,16 @@ protected function prepareInsertForDml($values)
...
-            return $this->connection->selectWithTimestampBound(
-                $this->toSql(), $this->getBindings(), $this->timestampBound
-            );
+            $options += $this->timestampBound->transactionOptions();
         }
 
-        return parent::runSelect();
+        return count($options) > 0
+            ? $this->connection->selectWithOptions($sql, $bindings, $options)
+            : $this->connection->select($sql, $bindings);
```
halnique, line 145:
```suggestion
        return $this->connection->selectWithOptions($sql, $bindings, $options); 
```
I think it would be simpler to just `selectWithOptions`.

src/Concerns/ManagesStaleReads.php
```diff
@@ -20,56 +20,45 @@
...
-     * @param  TimestampBoundInterface|null $timestampBound
+     * @deprecated use selectWithOptions() instead. This method will be removed in v7.
+     * @param string $query
+     * @param array<array-key, mixed> $bindings
+     * @param TimestampBoundInterface|null $timestampBound
      * @return list<array<array-key, mixed>|null>
      */
     public function selectWithTimestampBound($query, $bindings = [], TimestampBoundInterface $timestampBound = null): array
     {
-        return $this->withSessionNotFoundHandling(function () use ($query, $bindings, $timestampBound) {
```
halnique, line 57: Is `withSessionNotFoundHandling` no longer needed?

src/Connection.php
taka-oyama: iterator_to_array must happen within `run` or else certain error cannot be captured.

src/Concerns/ManagesStaleReads.php
halnique: thanks!

halnique (approved):

PR #124, opened 2023-07-31
feat: add method for adding row deletion policy (for modifying table)

taka-oyama:
Completely forgot that it can be added later.


tyabu12 (approved):
LGTM

PR #127, opened 2023-08-15
fix: escape list for Query/Builder::toRawSql

taka-oyama:
Spanner bindings can also include list values.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -37,6 +37,8 @@ use Google\Cloud\Spanner\Transaction;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Connection as BaseConnection;
 use Illuminate\Database\QueryException;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
 use InvalidArgumentException;
 use LogicException;
 use Psr\Cache\CacheItemPoolInterface;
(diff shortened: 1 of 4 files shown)


tests/ConnectionTest.php
```diff
@@ -482,4 +483,23 @@ protected function performRollBack($toLevel): void
...
+        self::assertSame('0', $conn->escape(0));
+        self::assertSame('-1', $conn->escape(-1));
+        self::assertSame('1.1', $conn->escape(1.1));
+        self::assertSame('"a"', $conn->escape('a'));
+        self::assertSame('r"""' . "\n" . '"""', $conn->escape("\n"));
+        self::assertSame('[]', $conn->escape([]));
+        self::assertSame('["a"]', $conn->escape(['a']));
+        self::assertSame('[false]', $conn->escape([false]));
+        self::assertSame('[1]', $conn->escape([1]));
+        self::assertSame('[1.1]', $conn->escape([1.1]));
```
halnique, line 503:
```suggestion
        self::assertSame('[1.1]', $conn->escape([1.1]));
        self::assertSame('[1, [2, [3]]]', $conn->escape([1, [2, [3]]]));
```

Shouldn't multidimensional arrays also be included in the assertion to test the circular process?

halnique (commented):
Also, please update `CHANGELOG.md`.

tests/ConnectionTest.php
taka-oyama: multi dimensional arrays are not supported.

https://cloud.google.com/spanner/docs/reference/standard-sql/data-types#declaring_an_array_type

halnique: Oops... It seems to pass the test case for multidimensional arrays, so I guess we need to fix it so it doesn't pass?

taka-oyama: That should be the expected behavior. This has been fixed.

src/Connection.php
```diff
@@ -37,6 +37,8 @@
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Connection as BaseConnection;
 use Illuminate\Database\QueryException;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
```
halnique, line 41:
```suggestion
```

Probably forgot to erase it.

halnique (approved):

PR #152, opened 2023-12-08
fix: separate auth cache and session pool to prevent race condition

taka-oyama:
Coming from https://github.com/googleapis/google-cloud-php/issues/6703#issuecomment-1846328959

Auth does not lock the cache file, so it's possible for a race condition to occur.

Below is an example of how that may happen.

1. Process 1: Auth opens cache
2. Process 2: Session Pool opens cache
3. Process 2: Session Pool creates a session and saves (file is written)
4. Process 1: Auth writes to cache (file is overwritten and session created at 3. no longer exists in cache)

- [x] CHANGELOG

Must be backported to 6.0.

### DIFF
--- a/src/SpannerServiceProvider.php
+++ b/src/SpannerServiceProvider.php
@@ -21,10 +21,12 @@ use Colopl\Spanner\Console\CooldownCommand;
 use Colopl\Spanner\Console\SessionsCommand;
 use Colopl\Spanner\Console\WarmupCommand;
 use Google\Cloud\Spanner\Session\CacheSessionPool;
+use Google\Cloud\Spanner\Session\SessionPoolInterface;
 use Illuminate\Database\DatabaseManager;
 use Illuminate\Queue\QueueManager;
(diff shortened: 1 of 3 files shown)


src/SpannerServiceProvider.php
```diff
@@ -56,40 +58,72 @@ public function boot(): void
...
+     *       cache_path: string,
+     *       session_pool: array<string, mixed>,
+     *  } $config
      */
     protected function parseConfig(array $config, string $name): array
     {
         return $config + [
             'prefix' => '',
             'name' => $name,
+            'cache_path' => null,
```
halnique, line 100:
```suggestion
            'cache_path' => '',
```
If it returns null, you may want to change the type hinting as well.

taka-oyama: type hint changed.

halnique (approved):

t-matsuno-777 (approved):

Issue #149, opened 2023-11-28
JSON columns dont appear to be working


matthewjumpsoffbuildings:
I have a model with a json column. As usual I have `$casts = ['json_column' => 'array']` in the model definition, but for some reason every time I try to save/update the model I get this error

```
Value of type STRING cannot be assigned to data, which has type JSON 
```

taka-oyama:
There is no good solution to this. 
The query builder has no way of distinguishing between a normal string and a json string.

taka-oyama:
Spanner didn't have JSON types initially (2019) and we just stored JSON as STRING(MAX).
That was good enough for our use case.

I'll update the docs to say it's not supported for the moment.

taka-oyama:
PR is always welcome btw.

matthewjumpsoffbuildings:
I haven't dealt with custom database/eloquent drivers before so its hard to follow the chain of how data is set/escaped before being passed to the `execute` statement.

I am trying to work out how the `escape` functions work here:
https://github.com/colopl/laravel-spanner/blob/724159abe1ee844fd4cec7d16c08c50d4d200d7e/src/Connection.php#L419C3-L442C6

I was thinking perhaps if you pass an associative array, the `escapeArray()` method, instead of throwing an exception, could treat it as JSON? 

But in my experimenting it doesnt even seem that the `escapeArray()` method gets called, at least not before this
https://github.com/colopl/laravel-spanner/blob/724159abe1ee844fd4cec7d16c08c50d4d200d7e/src/Connection.php#L319

taka-oyama:
> I am trying to work out how the escape functions work here

From what I can recall, `escape` is only used for the `Builder::toRawSql` which is only really used for logging and debugging.

taka-oyama:
> I was thinking perhaps if you pass an associative array, the escapeArray() method, instead of throwing an exception, could treat it as JSON?

That could potentially work but you might need to pass it as an object since empty arrays can't be treated as assoc array.

matthewjumpsoffbuildings:
The solution is a custom Cast with a custom ValueInterface, as per this comment

https://github.com/colopl/laravel-spanner/pull/154#issuecomment-1849330874

Should add this to the docs

matthewjumpsoffbuildings:
PR #265 sorts this

PR #154, opened 2023-12-09
add functionality to pass spanner types to insert/update

matthewjumpsoffbuildings:
This is a first attempt to resolve https://github.com/colopl/laravel-spanner/issues/149

The first step is to add a `$types` array to both Models and Builders, and to automatically inject a Models types array into every new Builder instance for that model.

The next step is overriding the base update/insert methods of the Builder, to loop through each value being passed in, and using a function `checkForType()` that attempts to apply the same conventions as the Parameterizer class, to generate an array of bound types eg `['p2' => Database::TYPE, ...]`.

The final step is adding an additional optional `$types = []` argument to Connection `update` / `insert` / `statement` / `affectingStatement` methods, so that it can be passed to the `$transaction->executeUpdate()` as an additional `types` argument alongside `parameters`

I am still not super familiar with how all the internals work here, likely there are several improvements that could be made, but hopefully this is a good jumping off point

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -280,7 +280,7 @@ class Connection extends BaseConnection
     /**
      * @inheritDoc
      */
-    public function statement($query, $bindings = [], $types = []): bool
+    public function statement($query, $bindings = [], array $types = []): bool
     {
         // is SELECT query
         if (0 === stripos(ltrim($query), 'select')) {
@@ -298,19 +298,26 @@ class Connection extends BaseConnection
         return $this->runDdlBatch([$query]) !== null;
     }
 
-    public function insert($query, $bindings = [], $types = [])
+    /**
+     * @inheritDoc
+     */
+    public function insert($query, $bindings = [], array $types = []): bool
     {
         return $this->statement($query, $bindings, $types);
     }
 
-    public function update($query, $bindings = [], $types = [])
-    {
-        return $this->affectingStatement($query, $bindings, $types);
-    }
     /**
      * @inheritDoc
      */
-    public function affectingStatement($query, $bindings = [], $types = []): int
+    public function update($query, $bindings = [], array $types = []): int
+    {
+        return $this->affectingStatement($query, $bindings, $types);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function affectingStatement($query, $bindings = [], array $types = []): int
     {
         /** @var Closure(): int $runQueryCall */
         $runQueryCall = function () use ($query, $bindings, $types) {
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -280,7 +280,7 @@ class Connection extends BaseConnection
     /**
      * @inheritDoc
      */
-    public function statement($query, $bindings = []): bool
+    public function statement($query, $bindings = [], $types = []): bool
     {
         // is SELECT query
         if (0 === stripos(ltrim($query), 'select')) {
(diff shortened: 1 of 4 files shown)


matthewjumpsoffbuildings:
Theres some issues with the types of some of the methods coming up in the unit tests, hopefully if the general ideas of this PR work, you dont mind sorting those, im not quite sure how to resolve them

taka-oyama:
I was looking through the underlying implementation and it looks like it would be easier if you implement a `Json` class which implements [ValueInterface](https://github.com/googleapis/google-cloud-php/blob/e04603e163e33578cec39a056017afc4f2e91588/Spanner/src/ValueInterface.php) and pass that as binging.

Coversion takes place [here](https://github.com/googleapis/google-cloud-php/blob/e04603e163e33578cec39a056017afc4f2e91588/Spanner/src/ValueMapper.php#L738-L746).

matthewjumpsoffbuildings:
Right so perhaps I could create a custom Cast class that uses the ValueInterface? That would be much simpler, this was a very hacky attempt

matthewjumpsoffbuildings:
Though I notice this in the comments:
```
* Note: This interface may become internal in the next major version change
```
Which may be concerning?

taka-oyama:
In that case I think it's best to throw a PR to them to add a JSON type so it's handled internally.
Seems like they already have one for [PgJsonb](https://github.com/googleapis/google-cloud-php/blob/e04603e163e33578cec39a056017afc4f2e91588/Spanner/src/PgJsonb.php).

matthewjumpsoffbuildings:
Just checking, do I use these google type interfaces anyway? Am I on the right track to assume its via Casts?

Also I need JSON working yesterday haha, I am not sure how long a PR will take with google

matthewjumpsoffbuildings:
Are you familiar with the ValueMapper class? Apparently this is a thing:
```php
$jsonValue = new PgJsonb('{"key": "value"}');

$valueMapper = new ValueMapper($db, true);

$mappedValue = $valueMapper->encode($jsonValue);

$result = $db->execute('INSERT INTO MyTable (JsonColumn) VALUES (@jsonValue)', [
   'parameters' => [
       'jsonValue' => $mappedValue
   ]
]);
```

taka-oyama:
> Just checking, how do I use these google type interfaces anyway? Am I on the right track to assume its via Casts?

I had to do this for Numeric types. You will have to convert to a ValueInterface when inserting/updating by overriding [Model::getDirtryForUpdate()](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L2046) and [Model::getAttributesForInsert](https://github.com/laravel/framework/blob/0d5072b9ebb4f1de2be73f28e548b7271c75733c/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1803).

matthewjumpsoffbuildings:
Holy moly i figured it out and its even easier - no changes required to your code at all.

If I make a custom cast like so:

```php
namespace App\Casts;

use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
use Google\Cloud\Spanner\PgJsonb;
use Google\Cloud\Spanner\V1\TypeAnnotationCode;

class SpannerJsonCast implements CastsAttributes
{
	public function get($model, $key, $value, $attributes)
	{
		return json_decode($attributes[$key], true);
	}

	public function set($model, $key, $value, $attributes)
	{
		return [$key => new SpannerJsonValue($value)];
	}
}

// Just copy the PgJsonb ValueInterface class
class SpannerJsonValue extends PgJsonb
{
	// override the PgJsonb classes typeAnnotation TypeAnnotationCode::PG_JSONB
	public function typeAnnotation()
	{
		return TypeAnnotationCode::TYPE_ANNOTATION_CODE_UNSPECIFIED;
	}
}
```

Then in my model I use that cast like so:

```php
	protected $casts = [
		'ColumnName' => SpannerJsonCast::class,
	];
```

taka-oyama:
> it throws an error from spanner saying it cant write strings to int columns

Doesn't happen to me. 
If you pass a form input, which is a string, this will happen. You have to cast it to an int first.

>  I am wondering where the implementation for your Numeric type solution is? Or has it not been released?

This is not released to public because it's not a very clean solution.

matthewjumpsoffbuildings:
> Doesn't happen to me. If you pass a form input, which is a string, this will happen. You have to cast it to an int first like `$model->number = (int) $value`.

I am using an Eloquent model, with `$casts = [ 'column' => 'int']`, and the form is actually a Laravel Nova form, using a Number field. None of this works, always seem to get the 'cannot insert string into int64' error from spanner. The only thing that did work was making a custom cast class that uses an extension of the Numeric value type

taka-oyama:
Then just override [Model::getDirtryForUpdate](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L2046) and [Model::getAttributesForInsert](https://github.com/laravel/framework/blob/0d5072b9ebb4f1de2be73f28e548b7271c75733c/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1803) and cast it to an int.

taka-oyama:
Also please provide as much context as possible so we don't have to go back and forth.

taka-oyama:
Custom type casting will be improved once google implements https://github.com/googleapis/google-cloud-php/issues/5838. Closing this for now.

PR #156, opened 2023-12-12
query grammer compileLock - ignore

matthewjumpsoffbuildings:
While its true spanner doesnt support explicit locking, I think the compileLock method should not throw an exception. It may be better to log a warning perhaps, but since spanner has complex lock handling and consistency, just gracefully allowing the lock to generate no additional code seems acceptable.

When using spanner with other frameworks (Laravel Nova in this case), it is necessary to ignore this.

Also if you look at: 
https://github.com/laravel/framework/blob/7907bfb50165b3951403e2d9882b4f8e0ee26b9e/src/Illuminate/Database/Query/Grammars/SQLiteGrammar.php#L29

and:
https://github.com/laravel/framework/blob/7907bfb50165b3951403e2d9882b4f8e0ee26b9e/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php#L340

You will see both of those return empty strings - I think graceful failure here is the much better approach

### DIFF
--- a/src/Query/Grammar.php
+++ b/src/Query/Grammar.php
@@ -43,7 +43,7 @@ class Grammar extends BaseGrammar
      */
     protected function compileLock(Builder $query, $value)
     {
-        $this->markAsNotSupported('explicit locking');
+        return '';
     }
 
     /**
--- a/tests/Query/BuilderTest.php
+++ b/tests/Query/BuilderTest.php
@@ -792,32 +792,26 @@ class BuilderTest extends TestCase
 
     public function test_lock(): void
     {
-        $this->expectException(BadMethodCallException::class);
-        $this->expectExceptionMessage('Cloud Spanner does not support explicit locking');
-
         $conn = $this->getDefaultConnection();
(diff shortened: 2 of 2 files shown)


taka-oyama:
Please add tests for this as well.

taka-oyama:
I think fixing the ones that are failing should be enough for this case.

matthewjumpsoffbuildings:
I see the 3 tests in BuilderTest.php, test_lock, test_lockForUpdate, test_sharedLock. Currently they depend on an exception being thrown, given the compileLock is now just returning an empty string what is the expected result?

taka-oyama:
Please change it so that `toSql()` produces the expected query.

matthewjumpsoffbuildings:
So drop those 3  test_lock, test_lockForUpdate, test_sharedLock methods in BuilderTest.php and just add to the test_toRawSql method?

matthewjumpsoffbuildings:
updated the 3 lock unit test methods

taka-oyama (approved):

taka-oyama:
Thanks.

Issue #133, opened 2023-08-25
Support auto generating primary key defaults


taka-oyama:
https://cloud.google.com/spanner/docs/primary-key-default-value

Add something like `autoGenerate()` for INT64 and string columns

matthewjumpsoffbuildings:
This relates to https://github.com/colopl/laravel-spanner/issues/3

Ideally the `increments()` method of Schema builders would also use bit reversed sequences as discussed there

matthewjumpsoffbuildings:
Is there any movement on this?

taka-oyama:
Waiting for [spanner emulator](https://github.com/GoogleCloudPlatform/cloud-spanner-emulator) to implement it so I can test it.

matthewjumpsoffbuildings:
This has been released, as well as Postgres dialect support

https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/releases/tag/v1.5.12

taka-oyama:
Still waiting for https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/140.

PR #157, opened 2023-12-12
auto generate uuids for incrememnts() columns

matthewjumpsoffbuildings:
As discussed in this issue https://github.com/colopl/laravel-spanner/issues/133 - initially was thinking using a bit-reversed sequence for auto-incrementing ids would be the optimal solution, but really i think UUIDs are probably better anyway, and more in line with how spanner operates. For example, I believe the bit reversed sequence generator is not as performant as auto generated UUIDs at scale

Added tests, all pass on the latest Spanner emulator


taka-oyama:
> I believe the bit reversed sequence generator is not as performant as auto generated UUIDs at scale
Do you have anything to backup this claim?

matthewjumpsoffbuildings:
> > I believe the bit reversed sequence generator is not as performant as auto generated UUIDs at scale
> 
> Do you have anything to backup this claim?

I have had email and video call contact with some engineers on the google Spanner team, as I was having some issues at scale with load tests, and they suggested that. Also if you consider how the bit reversed sequences are generated, from [the docs
](https://cloud.google.com/spanner/docs/primary-key-default-value#bit-reversed-sequence)
> A bit-reversed sequence is a schema object that produces a sequence of integers and bit-reverses them. This object uses bit reversal on a private, internal Spanner counter to ensure uniqueness...
> In Spanner, you use SEQUENCE DDL statements along with the bit_reversed_positive attribute to create, alter, or drop a sequence that produces bit-reversed positive values...
> Each sequence maintains a set of internal counters and uses them to generate a value. The sequence counter provides the input to the bit-reversing algorithm.

Unlike UUIDs, which are not state dependent, and are just randomly generated by Spanner, bit-reversed sequences hold a set of counters which are incremented by Spanner internally, which can be less performant when experiencing a high volume of writes

taka-oyama:
Did you try the load test with `GENERATE_UUID`?

matthewjumpsoffbuildings:
Hmm the unit test is failing here, but on my local Spanner emulator it was running fine? I am guessing the Dockerfile or something needs updating to the latest Spanner?

matthewjumpsoffbuildings:
> Did you try the load test with `GENERATE_UUID`?

Yes and it was more performant. The most performant option is to generate the UUID client side, but still if one is going to use the auto-generated columns, I think UUID is a better choice

taka-oyama:
Dockerfile is fine. It pulled the latest emulator.

Please check the error. There is something wrong with your code.

```

1) Colopl\Spanner\Tests\Schema\BlueprintTest::test_increments
ErrorException: Undefined array key "large"

/project/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php:254
/project/tests/Schema/BlueprintTest.php:599

ERRORS!
Tests: 137, Assertions: 456, Errors: 1, Skipped: 2.
2
```

matthewjumpsoffbuildings:
Yes, my mistake, sorry, pushed a new commit

taka-oyama:
> if one is going to use the auto-generated columns, I think UUID is a better choice

That's a fair point. Especially if bit reversed sequence generator has a scaling issue.

src/Schema/Blueprint.php
```diff
@@ -18,8 +18,10 @@
 namespace Colopl\Spanner\Schema;
 
 use Colopl\Spanner\Concerns\MarksAsNotSupported;
+use Illuminate\Database\Query\Expression;
 use Illuminate\Database\Schema\Blueprint as BaseBlueprint;
 use Illuminate\Database\Schema\ColumnDefinition;
+use Illuminate\Support\Facades\DB;
```
taka-oyama, line 24: This doesn't look like it's used?

```suggestion
```

matthewjumpsoffbuildings:
> > if one is going to use the auto-generated columns, I think UUID is a better choice
> 
> That's a fair point. Especially if bit reversed sequence generator has a scaling issue.

I would say bit rev seq is still performant, just not as optimal as UUID

taka-oyama:
Closing since I have created a similar PR [here](https://github.com/colopl/laravel-spanner/pull/175) which solves the same problem.

PR #161, opened 2023-12-12
dropAllTables

matthewjumpsoffbuildings:
Update: doing a proper step by step drop of all tables/indexes/foreign keys

As the title says - a command to drop all tables might as well just drop the database and remake it, and this makes the `migrate:fresh` command work properly

I am not sure if I have accessed the connection the correct way, eg `DB::connection()`. Also theres no tests as of yet


taka-oyama:
Dropping all the tables by dropping the database is not ideal. This will drop all the backups and monitoring info such as system insight.

matthewjumpsoffbuildings:
Yea its far from ideal haha, the only other thing I can think to do is a bit of a hack (but it works) is to get the generated sql queries from all migrations `up()` methods, then convert all `create table` statements to `drop table`, all `create index` to `drop index` etc. You cant even use the `down()` methods since many internal Laravel migrations dont have any `dropForeign` or `dropIndex` statements, and annoyingly with Spanner you have to drop indexes and constraints first before dropping the table

taka-oyama:
How about getting all the indexes and tables from `Schema\Builder::getIndexListing(...)` and `Schema\Builder::getTables(...)` and just drop them?

matthewjumpsoffbuildings:
That seems much better, is there a similar method for foreign keys?

matthewjumpsoffbuildings:
Also I am noticing there is no method for getTables, there is getAllTables but that contains a deprecated call to  $this->grammar->compileGetAllTables()?

matthewjumpsoffbuildings:
Also I see there is no method to compile a foreign key listing, but I can make on in the Grammar

taka-oyama:
> Also I am noticing there is no method for getTables

https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Schema/Builder.php#L195-L200

matthewjumpsoffbuildings:
> > Also I am noticing there is no method for getTables
> 
> https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Schema/Builder.php#L195-L200

Ye sorry my mistake, IDE wasnt picking it up for some reason

taka-oyama:
Getting the tables might not be enough since spanner can have interleaved tables and they have to be deleted first.

So you might have to get the relationship information by calling `INFORMATION_SCHEMA.TABLES` directly.

matthewjumpsoffbuildings:
I could probably make a similar method to the index/foreign key listing methods right? As far as interleaved tables goes - is it the child or the parent I have to delete first?

taka-oyama:
I think you have to delete the index / foreign keys first, then the tables.
For interleaved tables, you have to delete the children first.
I think you will have to build a tree structure and delete depth first.

matthewjumpsoffbuildings:
Done and working well, theres a test which has indexes, foreign keys, and nested interleaving.

The dropAllTables method is a bit ugly, for example I had to drop foreign keys first before listing indexes since they get listed as indexes too and I dont think theres a way to tell which index is for a foreign key and which is just a normal index.

This also requires the PR #163 to allow dropping foreign keys properly

matthewjumpsoffbuildings:
Weird, the tests were working locally, but it seems like my dropAllTables test is (unsurprisingly) dropping all tables and then a different test is failing cause the Users table is gone?

matthewjumpsoffbuildings:
So I modified phpunit.xml to have 2 directories in the testsuite, the first uses the normal suffix Test.php, and the second uses TestLast.php. This allows any test cases that need to be run at the end to do so.

Schema/BuilderTest.php has been renamed to BuilderTestLast.php, so the dropAll test will happen right at the end

src/Schema/Grammar.php
```diff
@@ -93,6 +93,16 @@ public function compileIndexListing()
         return 'select index_name as `index_name` from information_schema.indexes where table_schema = \'\' and table_name = ?';
     }
 
+    /**
+     * Compile the query to determine the list of foreign keys.
+     *
+     * @return string
+     */
+    public function compileForeignListing()
```
taka-oyama, line 101: Would it be possible to change this to `compileForeignKeys` which will align it with [other database grammars](https://github.com/laravel/framework/blob/eb948833d513088249d73775586d8fe0ccf48158/src/Illuminate/Database/Schema/Grammars/SQLiteGrammar.php#L155)?

matthewjumpsoffbuildings: Yes, doing now, will push in 5

src/Schema/Builder.php
```diff
@@ -64,6 +79,24 @@ public function getIndexListing($table)
         return $processor->processIndexListing($results);
     }
 
+    /**
+     * @param string $table
+     * @return string[]
+     */
+    public function getForeignListing($table)
```
taka-oyama, line 86: Maybe you can use [getForeignKeys()](https://github.com/laravel/framework/blob/eb948833d513088249d73775586d8fe0ccf48158/src/Illuminate/Database/Schema/Builder.php#L376) instead?

matthewjumpsoffbuildings: I feel like in this file its more in line with the other method `getIndexListing()` just above?

matthewjumpsoffbuildings: But happy to change that too

src/Schema/Builder.php
```diff
@@ -19,7 +19,10 @@
 
 use Closure;
 use Colopl\Spanner\Query\Processor;
+use Colopl\Spanner\Connection;
 use Illuminate\Database\Schema\Builder as BaseBuilder;
+use Illuminate\Support\Facades\DB;
```
taka-oyama, line 24:
```suggestion
```

src/Schema/Builder.php
matthewjumpsoffbuildings: It also uses the `$processor->processIndexListing()` method internally, its basically a 1:1 copy of `getIndexListing()` so I thought I would keep the naming convention the same

taka-oyama: The internals of `getForeignKeys()` looks like the below, so maybe we can just implement `processForeignKeys` in the PostProcessor. That way it's more compatible with other databases.
```
$table = $this->connection->getTablePrefix().$table;

return $this->connection->getPostProcessor()->processForeignKeys(
    $this->connection->selectFromWriteConnection($this->grammar->compileForeignKeys($table))
);
```

matthewjumpsoffbuildings: If we want to be consistent with (https://github.com/laravel/framework/blob/eb948833d513088249d73775586d8fe0ccf48158/src/Illuminate/Database/Schema/Builder.php)

shouldnt `getIndexListing` be `getIndexes` too? And then the Processor should have `processIndexes` and `processForeignKeys`?

matthewjumpsoffbuildings: Regardless I changed getForeignKeyListing to getForeignKeys and added processForeignKeys, this PR should be ready to merge

taka-oyama: FYI `getIndexes` was [added to the framework last week](https://github.com/laravel/framework/pull/49204).

matthewjumpsoffbuildings: Ah makes sense

src/Schema/Builder.php
```diff
@@ -64,6 +78,24 @@ public function getIndexListing($table)
         return $processor->processIndexListing($results);
     }
 
+    /**
+     * @param string $table
+     * @return string[]
+     */
+    public function getForeignKeys($table)
```
taka-oyama, line 85: Is this code even necessary? It does the same thing as the parent method.

src/Schema/Builder.php
```diff
@@ -46,6 +48,18 @@ public function getAllTables()
         );
     }
 
+    /**
+     * @inheritDoc
+     *
+     * @return list<array{ name: string, type: string, parent: string }>
+     */
+    public function getTables()
```
taka-oyama, line 56: getTables should just work without overide?

src/Schema/Builder.php
matthewjumpsoffbuildings: Yea you may be right, will remove override and test

taka-oyama: Seems like the parent method was added 3 days ago and is only available in laravel/framework `v10.37.0`.
Can you bump the composer.json dependency to `10.37.3`?

src/Schema/Builder.php
taka-oyama: Nice. Thanks.

taka-oyama: Can you add a comment stating the difference from parent?

matthewjumpsoffbuildings:
Hmm that appears not to be working without the override

taka-oyama:
Seems like there's a binding missing?

```
1) Colopl\Spanner\Tests\Schema\BuilderTestLast::test_dropAllTables
ErrorException: Undefined array key 0

/project/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php:254
/project/src/Query/Parameterizer.php:47
/project/src/Query/Parameterizer.php:46
/project/src/Connection.php:468
/project/vendor/laravel/framework/src/Illuminate/Database/Connection.php:776
/project/src/Connection.php:262
/project/src/Connection.php:239
/project/vendor/laravel/framework/src/Illuminate/Database/Connection.php:394
/project/vendor/laravel/framework/src/Illuminate/Database/Schema/Builder.php:381
/project/src/Schema/Builder.php:153
/project/tests/Schema/BuilderTestLast.php:290
```

matthewjumpsoffbuildings:
Yes in reality the newer Laravel way of doing things required a `$table` to be passed to the compile function. I have now aligned `getForeignKeys($table)` and `compileForeignKeys($table)` with the newer approach

I have also aligned `getIndexes($table)` and `compileIndexes($table` with the  standard Laravel Schema Grammar and Builder methods

src/Schema/Builder.php
```diff
@@ -47,21 +49,15 @@ public function getAllTables()
     }
 
     /**
-     * @param string $table
-     * @return string[]
+     * @inheritDoc Adds a parent key, for tracking interleaving
+     * 
+     * @return list<array{ name: string, type: string, parent: string }>
      */
-    public function getIndexListing($table)
```
taka-oyama, line 53: Don't delete existing methods on minor updates, just deprecate them.

taka-oyama (approved):

PR #162, opened 2023-12-13
Allow pretending for DDL statements

matthewjumpsoffbuildings:
If the Connection is pretending, the DDL statements should not run

Since affectingStatement already handles pretending, its fine to just pass through to that method which will return 0 as normal, allowing the generated SQL to be captured


src/Schema/Builder.php
```diff
@@ -165,6 +166,8 @@ public function dropAllTables()
             $blueprint->drop();
             array_push($queries, ...$blueprint->toSql($connection, $this->grammar));
         }
-        $connection->runDdlBatch($queries);
+
+        if(count($queries))
+            $connection->runDdlBatch($queries);
```
taka-oyama, line 171: These fixes don't belong in this PR.

src/Connection.php
```diff
@@ -295,6 +295,10 @@ public function statement($query, $bindings = []): bool
         }
 
         // is DDL Query
+        // allow pretending for DDL queries
+        if($this->pretending())
+            return $this->affectingStatement($query, $bindings) !== null;
```
taka-oyama, line 300: I think runDdlBatch should be fixed to handle the pretending instead of forwarding it to affectingStatement.

src/Schema/Builder.php
matthewjumpsoffbuildings: ye fair i merged from the other drop-all branch into this

src/Connection.php
matthewjumpsoffbuildings: but isnt affectingStatement()s implementation of `pretend` is basically complete and simple and i would just end up duplicating affectingStatement() line for line, except for this range of lines

https://github.com/colopl/laravel-spanner/blob/master/src/Connection.php#L313-L323

matthewjumpsoffbuildings: or is nothing in the affectingStatement method relevant?

taka-oyama:
Fixed in https://github.com/colopl/laravel-spanner/pull/170.

PR #164, opened 2023-12-18
Drop all bugfix

matthewjumpsoffbuildings:
If there are no tables, or no indexes or foreign keys, the current drop all method errors since it attempts to runDdlBatch with an empty array

This PR fixes that


taka-oyama:
Can you add tests for this case? Thanks.

taka-oyama:
Fixed with https://github.com/colopl/laravel-spanner/pull/169.

PR #174, opened 2024-01-24
feat: add support for GENERATE_UUID

taka-oyama:
Fix: #173 

## Checklist

- [x] CHANGELOG


src/Schema/ColumnDefinition.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Colopl\Spanner\Schema;
+
+use Illuminate\Database\Query\Expression;
+use Illuminate\Database\Schema\ColumnDefinition as BaseColumnDefinition;
+
+class ColumnDefinition extends BaseColumnDefinition
+{
+    public function generateUuid(): static
```
tyabu12, line 10:
```suggestion
    /*
     * @return $this
     */
    public function generateUuid(): static
```

tyabu12 (changes requested):

src/Schema/Blueprint.php
```diff
@@ -94,7 +95,16 @@ public function tinyIncrements($column)
...
     }
 
-    // region Spanner Specific Types
+    /**
+     * @inheritDoc
+     * @return SpannerColumnDefinition
+     */
+    public function uuid($column = 'uuid')
+    {
+        $definition = new SpannerColumnDefinition(['type' => 'uuid', 'name' => $column]);
```
tyabu12, line 104: How about overriding `addColumn()` method for consistency?

taka-oyama: That would make it more consistent but less robust. `generate_uuid` should only be available when a column is defined as uuid. I've renamed `ColumnDefinition` to `UuidColumnDefinition` for clarity.

tyabu12 (approved):
LGTM

taka-oyama:
Thanks for the review!

PR #178, opened 2024-01-26
chore: remove deprecated methods

taka-oyama:
- `Connection::runDdl()` and `Connection::runDdls()` has been removed. Use `Connection::runDdlBatch()` instead.
- Trait `ManagesStaleReads` has been removed (which contained `cursorWithTimestampBound()` and `selectWithTimestampBound()`. Use `selectWithOptions()` instead).
- `Blueprint::interleave()` and `IndexDefinition::interleave()` now throw an error instead of a deprecation.


src/Concerns/ManagesDataDefinitions.php
```diff
@@ -30,28 +30,6 @@ trait ManagesDataDefinitions
...
     abstract public function getSpannerDatabase(): Database;
 
-    /**
-     * @deprecated use runDdlBatch() instead
-     * @param string $ddl
-     * @return LongRunningOperation
-     */
-    public function runDdl(string $ddl): LongRunningOperation
-    {
-        trigger_deprecation('colopl/laravel-spanner', '5.2', 'runDdl() is deprecated. Use runDdlBatch() instead.');
```
tyabu12, line 40: `use function trigger_deprecation;` is no longer used.

tyabu12 (changes requested):

src/Concerns/ManagesDataDefinitions.php
taka-oyama: deleted.

tyabu12 (approved):
LGTM

taka-oyama:
Thanks for reviewing

PR #180, opened 2024-02-15
feat: allow spanner_emulator.disable_query_null_filtered_index_check to be set

taka-oyama:
An [error is thrown](https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/blob/123682157cbf8fa5cb13f0d7edb9b66070bb4a11/common/errors.cc#L2539-L2554) when the following condition is met.

- Spanner Emulator is used
- index is defined as `NULL_FILTERED`
- FORCE_INDEX statement hint is set on the query

This allows it to be bypassed by calling `Builder::disableEmulatorNullFilterIndexCheck()`.

### DIFF
--- a/src/Query/Builder.php
+++ b/src/Query/Builder.php
@@ -21,6 +21,8 @@ use Colopl\Spanner\Connection;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Query\Builder as BaseBuilder;
 use Illuminate\Support\Arr;
+use LogicException;
+use Override;
 
 class Builder extends BaseBuilder
 {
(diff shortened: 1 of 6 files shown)


src/Query/Builder.php
```diff
@@ -147,4 +149,29 @@ protected function runSelect()
...
+     * @return $this
+     */
+    public function forceIndex($index): static
+    {
+        $this->indexHint = new IndexHint('force', $index);
+
+        return $this;
+    }
+
+    public function disableEmulatorNullFilterIndexCheck(): static
```
tyabu12, line 164:
```suggestion
    /**
     * @return $this
     */
    public function disableEmulatorNullFilterIndexCheck(): static
```

tyabu12 (changes requested):

src/Query/Builder.php
yamadashy, line 164: 細かいのですが、 disableEmulatorNullFilter **ed** IndexCheck() にしていないのは意図されていますでしょうか？

taka-oyama: fixed

src/Query/Builder.php
```diff
@@ -21,6 +21,8 @@
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Query\Builder as BaseBuilder;
 use Illuminate\Support\Arr;
+use LogicException;
+use Override;
```
yamadashy, line 25:
```suggestion
```
使ってなさそうです。

yamadashy (approved):

tyabu12 (approved):
LGTM

PR #181, opened 2024-02-20
feat: add support for SEQUENCE

taka-oyama:
Closes: #133


src/Schema/Blueprint.php
```diff
@@ -32,6 +31,26 @@ class Blueprint extends BaseBlueprint
 {
     use MarksAsNotSupported;
 
+    /**
+     * @inheritDoc
+     * @return IntColumnDefinition
+     */
+    public function bigInteger($column, $autoIncrement = false, $unsigned = false): IntColumnDefinition
```
tyabu12, line 38: Is handling `$autoIncrement` and `$unsigned` unnecessary?

It would seem helpful to call `markAsNotSupported()`.

src/Schema/Blueprint.php
```diff
@@ -296,6 +315,57 @@ public function dropRowDeletionPolicy(): Fluent
...
+        $definition = new SequenceDefinition($name);
+        $this->addCommand('createSequence', ['definition' => $definition, 'ifNotExists' => false]);
+        return $definition;
+    }
+
+    /**
+     * @param string $name
+     * @return SequenceDefinition
+     */
+    public function createSequenceIfNotExist(string $name): SequenceDefinition
```
tyabu12, line 333: This method is the only return type of `SequenceDefinition`.  
Is this return type correct as intended?

src/Schema/IntColumnDefinition.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
```
tyabu12, line 3: May I know the case of `declare(strict_types=1);` just for my information?

It is defined in `IntColumnDefinition` and not in `SequenceDefinition`.

src/Schema/Blueprint.php
```diff
@@ -296,6 +315,57 @@ public function dropRowDeletionPolicy(): Fluent
         return $this->addCommand('dropRowDeletionPolicy');
     }
 
+    /**
+     * @param string $name
+     * @return SequenceDefinition
```
tyabu12, line 320: Perhaps `Fluent<string, mixed>` would be more appropriate?

tyabu12 (changes requested):

src/Schema/IntColumnDefinition.php
taka-oyama: Newly added files have them. I forgot to add it to SequenceDefinition. Adding it now.

src/Schema/Blueprint.php
taka-oyama: SequenceDefinition is correct here, it was just not passed correctly, fixed.

src/Schema/Blueprint.php
taka-oyama: Would be nice but I think that's out of scope from this PR.

src/Schema/Blueprint.php
```diff
@@ -295,6 +315,54 @@ public function dropRowDeletionPolicy(): Fluent
...
+        return $command;
+    }
+
+    /**
+     * @param string $name
+     * @return Fluent<string, mixed>
+     */
+    public function dropSequence(string $name): Fluent
+    {
+        return $this->addCommand(__FUNCTION__, ['sequence' => $name, 'ifExists' => false]);
```
tyabu12, line 354: ifExists seems to be not used in compileDropSequence / compileDropSequenceIfExists ?

taka-oyama: Thanks. fixed.

README.md
```diff
@@ -265,6 +266,35 @@ $schemaBuilder->table('user', function (Blueprint $table) {
...
+    $table->integer('id')->useSequence();
+});
+```
+
+If you want to be more flexibility, you can also create, alter, and drop sequences directly as below.
+
+```php
+$schemaBuilder->create('user_items', function (Blueprint $table) {
+    $table->createSequence('sequence_name');
+    $table->integer('id')->useSequence('sequence_name');
```
tyabu12, line 286: There seems to be no test case for this. Can you add this test case?

src/Schema/Blueprint.php
taka-oyama: This is no longer the case, all except dropXxx return SequenceDefinition now.

README.md
taka-oyama: Test added.

src/Schema/Grammar.php
```diff
@@ -631,6 +697,19 @@ protected function typeBoolean(Fluent $column)
...
 
+    /**
+     * @param Blueprint $blueprint
+     * @param Fluent<string, mixed> $column
+     * @return string|null
+     */
+    protected function modifyUseSequence(Blueprint $blueprint, Fluent $column): ?string
+    {
+        if (isset($column->useSequence)) {
+            return " default (get_next_sequence_value(sequence {$column->useSequence}))";
```
tyabu12, line 708: It would be better to wrap like other methods?

```suggestion
            return " default (get_next_sequence_value(sequence {$this->wrap($column->useSequence)}))";
```

taka-oyama: Fixed.

tyabu12 (approved):
LGTM!

PR #183, opened 2024-02-20
chore: fix formatting and spelling errors

taka-oyama:


tests/Query/DatetimeTest.php
```diff
@@ -55,19 +58,19 @@ public function testTimezoneWithQueryBuilder(): void
         $conn = $this->getDefaultConnection();
 
         date_default_timezone_set('Asia/Tokyo');
-        $expected = new \DateTime('2018-03-13 09:00:00');
+        $expected = new DateTime('2018-03-13 09:00:00');
         $this->assertEquals('Asia/Tokyo', $expected->getTimezone()->getName());
 
         $row = $conn->query()->selectRaw('TIMESTAMP("2018-03-13T00:00:00Z")')->get()->first();
-        /** @var \DateTimeInterface $datetime */
+        /** @var DateTimeInterface $datetime */
```
tyabu12, line 65: I think `$this->assertInstanceOf(DateTimeInterface::class, $datetime)` is better than PHPDoc overwrite in view of reliability.

taka-oyama: That change goes against the title so it will be done in a different PR.

tyabu12 (approved):
LGTM

tests/Query/DatetimeTest.php
taka-oyama: I think this makes sense but PHPDoc cannot be deleted because phpstan will throw an error here without it.

PR #185, opened 2024-02-20
chore: use better test methods

taka-oyama:
- Use `assertSame` instead of `assertEquals` (strict comparison vs loose comparison)
- `$this` instead of `self::` (some methods only exist in $this, so might as well be consistent)


tyabu12 (approved):
LGTM

PR #187, opened 2024-02-21
chore: partitionedDml can be tested in emulator now

taka-oyama:


tests/Query/BuilderTest.php
```diff
@@ -749,49 +749,30 @@ public function testWhereIn(): void
...
             $insertValues[] = $this->generateTestRow();
         }
+        $conn->table($tableName)->insert($insertValues);
 
-        collect($insertValues)->chunk(50)->each(function(Collection $chunk) use($conn, $tableName) {
-            $insertValues = array_values($chunk->all());
-            $conn->table($tableName)->insert($insertValues);
-        });
-
-        // normal DML should throw error since its over 20000 rows
```
taka-oyama, line 771: DML mutation limit has been raised to 80k and PHP now runs of out memory before it reaches that point so removing test to check the mutation limit.

tyabu12 (approved):
LGTM

PR #188, opened 2024-02-21
fix: prepareBinding should not have changed return type

taka-oyama:
Reverts part that changed the return type in https://github.com/colopl/laravel-spanner/pull/168.

Broke toRawSql formatting output for DateTime objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v7.0.1 (2024-02-21)
+# v7.0.1 (2024-02-22)
 
 Fixed
 - `Connection::prepareBinding()` should not have changed return type (#188)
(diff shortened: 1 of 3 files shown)


tyabu12 (approved):
LTGM

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v7.0.1 (2024-02-21)
```
taka-oyama, line 1:
```suggestion
# v7.0.1 (2024-02-22)
```

taka-oyama:
If this is applied, ARRAY<TIMESTAMP> will break so there's not clear answer to this. 
We'll just have to live with this.

taka-oyama:
CHANGELOG has been updated to address this.

https://github.com/colopl/laravel-spanner/blob/29a10c9e0aa0d168272eacd7f322a90c83372837/CHANGELOG.md?plain=1#L17

PR #193, opened 2024-03-04
dropAll() return immediately if no tables found

matthewjumpsoffbuildings:
Small tweak to Schema Builder dropAll() - if no tables exists simply return immediately

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v7.0.1? (?)
+# v7.1.0 (Not released yet)
 
 Changed
 - `Schema\Builder::dropAllTables` returns immediately if no tables exist (#193)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # v7.0.1? (?)
 
 Changed
-- `Schema\Builder::dropAll()` returns immediately if no tables exist (#193)
+- `Schema\Builder::dropAllTables` returns immediately if no tables exist (#193)
 
 # v7.0.0 (2024-02-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v7.0.1? (?)
+
+Changed
+- `Schema\Builder::dropAll()` returns immediately if no tables exist (#193)
+
 # v7.0.0 (2024-02-21)
 
 Added
--- a/tests/Schema/BuilderTestLast.php
+++ b/tests/Schema/BuilderTestLast.php
@@ -307,4 +307,18 @@ class BuilderTestLast extends TestCase
         $tables = $sb->getTables();
         $this->assertEmpty($tables);
     }
+
+    public function test_dropAllTablesEmpty(): void
+    {
+        $conn = $this->getDefaultConnection();
+        $sb = $conn->getSchemaBuilder();
+
+        $tables = $sb->getTables();
+        $this->assertEmpty($tables);
+
+        $sb->dropAllTables();
+
+        $tables = $sb->getTables();
+        $this->assertEmpty($tables);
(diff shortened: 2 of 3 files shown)


taka-oyama:
Please add a test that runs this code.

matthewjumpsoffbuildings:
Im actually not sure how to test this? Since dropAll() returns void, how would I test if it exited early due to there being no tables, or if it exited later because it found and dropped some tables?

taka-oyama:
I'm ok as long as all the code paths are covered.
So you can just run Connection::dropAllTables with no tables and just `$this->assertTrue(true);`.

matthewjumpsoffbuildings:
added test_dropAllTablesEmpty(), that should cover it?

matthewjumpsoffbuildings:
It appears to work when I run it, but I am not sure about a couple of things

Does that test_dropAllTablesEmpty() get run last in the file? If so it will be okay since the previous test_dropAllTables() will have emptied the db. Or do the test methods in the file get called in any order, in which case does each test method start with an empty database, or do artefacts from previous tests persist?

taka-oyama:
I think tests that depend on other tests to be run first is really bad.

So ideally, there should be a hook which deletes all created tables after every test, but that slows down the test significantly.
So as a compromise, I think you should drop all tables twice in this test. That way the 2nd try will always go through the code provided.

matthewjumpsoffbuildings:
Done

tests/Schema/BuilderTestLast.php
```diff
@@ -307,4 +307,19 @@ public function test_dropAllTables(): void
         $tables = $sb->getTables();
         $this->assertEmpty($tables);
     }
+
+    public function test_dropAllTablesEmpty(): void
+    {
+        $conn = $this->getDefaultConnection();
+        $sb = $conn->getSchemaBuilder();
+        $sb->dropAllTables();
```
taka-oyama, line 315:
```suggestion

         // All tables must be dropped before hand to ensure that this test will cover the "early return" case
         // for when there are no tables.
        $sb->dropAllTables();
```

tests/Schema/BuilderTestLast.php
```diff
@@ -307,4 +307,22 @@ public function test_dropAllTables(): void
         $tables = $sb->getTables();
         $this->assertEmpty($tables);
     }
+
+    public function test_dropAllTablesEmpty(): void
```
taka-oyama, line 311:
```suggestion
    public function test_dropAllTables_when_no_tables_exist(): void
```

taka-oyama (approved):

taka-oyama:
Please add a changelog

matthewjumpsoffbuildings:
Updated the changelog, not sure if its a new point release, please edit if needed

Also should we add #191 and #192 to the changelog too?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v7.0.1? (?)
```
taka-oyama, line 1:
```suggestion
# v7.1.0 (Not released yet)
```

taka-oyama:
> Also should we add https://github.com/colopl/laravel-spanner/pull/191 and https://github.com/colopl/laravel-spanner/pull/192 to the changelog too?

Yes but I'd like to have a different PR for them.

taka-oyama (approved):

matthewjumpsoffbuildings:
> Yes but I'd like to have a different PR for them.

Do you want me to do anything or you got it?

taka-oyama:
I'd appreciate it if you can write the ones for the previous two. Thanks.

matthewjumpsoffbuildings:
Which branch should I do it on?

taka-oyama:
master for all unreleased features.

PR #204, opened 2024-03-31
Spanner migrate commands

matthewjumpsoffbuildings:
Running migrations with spanner is very slow. This PR adds a couple of console commands

- `spanner:migrate`
- `spanner:migrate:fresh`

If you are using spanner, it will run the migrations in pretend mode, capture the output queries and convert them into batch DDL statements (and run any additional DML statements the migrations generate), resulting in massive improvements to migration duration

Also allows passing the `--seed` option as per the usual migrate commands, which calls `db:seed` when done

If you are not using spanner it falls back to using the normal `migrate` and `migrate:fresh` commands.

Added tests for both versions of the command, both with and without the `--seed` option

### DIFF
--- a/src/Console/MigrateCommand.php
+++ b/src/Console/MigrateCommand.php
@@ -71,7 +71,7 @@ class MigrateCommand extends Command
                 $repository->createRepository();
             $migrator->runPending($migrations);
         });
-    
+
         foreach($results as $result) {
             $query = $result['query'];
             if(Str::startsWith($query, 'select')) continue;
--- a/src/Console/MigrateCommand.php
+++ b/src/Console/MigrateCommand.php
@@ -14,13 +14,13 @@ class MigrateCommand extends Command
 {
     protected $signature = 'spanner:migrate {--seed}';
 
-    public function handle($fresh = false)
+    public function handle(bool $fresh = false):void
     {
         $connection = DB::connection();
 
         // check if we not in spanner mode
         if(!$connection instanceof Connection) {
-            $command = 'migrate' . $fresh ? ':fresh' : '';
+            $command = 'migrate' . ($fresh ? ':fresh' : '');
             $this->call($command, ['--seed' => $this->option('seed')]);
             return;
         }
@@ -28,8 +28,8 @@ class MigrateCommand extends Command
         // if running emulator, ensure instance exists
         if (!empty(getenv('SPANNER_EMULATOR_HOST'))) {
             $this->info('Checking Emulator Instance');
-            $spanner = new SpannerClient($connection->getConfig('client'));
-            $instanceName = $connection->getConfig('instance');
+            $spanner = new SpannerClient((array)$connection->getConfig('client'));
+            $instanceName = (string)$connection->getConfig('instance');
             if (! $spanner->instance($instanceName)->exists()) {
                 $config = $spanner->instanceConfiguration('emulator-config');
                 $spanner->createInstance($config, $instanceName)->pollUntilComplete();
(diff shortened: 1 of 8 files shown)


matthewjumpsoffbuildings:
Hmm the test is failing with this error
```
1) Colopl\Spanner\Tests\Console\SessionsCommandTest::test_no_args
Output does not contain "main contains 1 session(s).".

/project/vendor/laravel/framework/src/Illuminate/Testing/PendingCommand.php:354
/project/vendor/laravel/framework/src/Illuminate/Testing/PendingCommand.php:320
/project/tests/Console/SessionsCommandTest.php:59
```

I dont think this has anything to do with the new code?

matthewjumpsoffbuildings:
@taka-oyama if you think this is useful to merge, I can update the docs and changelog

src/Console/MigrateCommand.php
```diff
@@ -0,0 +1,103 @@
...
+class MigrateCommand extends Command
+{
+    protected $signature = 'spanner:migrate {--seed}';
+
+    public function handle(bool $fresh = false):void
+    {
+        $connection = DB::connection();
+
+        // check if we not in spanner mode
+        if(!$connection instanceof Connection) {
```
taka-oyama, line 22: You can specify multiple databases in a single migration file so checking connection this way is not sufficient.

src/Console/MigrateCommand.php
```diff
@@ -0,0 +1,103 @@
...
+        if(count($ddl)) {
+            $this->info('Running batch DDL');
+            $connection->runDdlBatch($ddl);
+        } else
+            $this->info('Nothing to migrate');
+
+        if(count($dml)) {
+            $this->info('Running DML');
+            foreach ($dml as $d) {
+                $connection->statement($d['query'], $d['bindings']);
```
taka-oyama, line 94: You can't just run DML's after, DDL's. In a case where DDL that runs after a DML might drop or alter the column you are trying to modify which will result in an error.

matthewjumpsoffbuildings: This isnt an issue since `alter` and `drop` statements are DDL


https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#alter_table_add_column_statement

https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#drop_table_statement

If you look at how `$dml` and `$ddl` is defined

```php
 foreach($results as $result) {
            $query = $result['query'];
            if(Str::startsWith($query, 'select')) continue;

            if(Str::startsWith($query, ['update', 'insert'])) 
                $dml[] = $result;
            else
                $ddl[] = $query;
        }
```

You will notice that I only add `insert` or `update` statements to `$dml` - any `alter` or `drop` statements will be run first, since they are added to `$ddl`

matthewjumpsoffbuildings: From what I understand, the only `insert` statements that the migrations generate is inserts into the `migrations` table?

What other insert/update statements would a migration make? Migrations deal with defining tables/columns/indexes, not inserting data, that is for the seeders

src/Console/MigrateCommand.php
matthewjumpsoffbuildings: Are you talking about this kind of thing
```php
Schema::connection('mysql')->create('test2', function(Blueprint $table)
        {
            $table->increments('id')->unsigned();
        });
```

Or the fact you can specify a database in the `migrate` console command
```
php artisan migrate --database="nameOfConnection"
```

matthewjumpsoffbuildings: I think for simple uses where you just have one database in the project, for now this is still handy. If its documented that for now it only works if you are using a single connection/database, with plans to improve later, its worth adding to the repo, no?

Dealing with multiple databases seems like it will be tricky to implement, but I will look into it

taka-oyama:
Stale.

PR #208, opened 2024-05-13
feat: allow request timeout at query level

taka-oyama:
Sometimes needed for "data boosted" queries since some requests take a long time to process.

- [ ] Backport to v7.


taka-oyama:
backported as v7.4.0.

Issue #209, opened 2024-05-19
Switching connections doesnt work due to AuthCache


matthewjumpsoffbuildings:
If I have 2 different spanner connections setup in my `config/databases.php`, with different project IDs and key files, if I connect to  one via `DB::connection('spanner-one')`, do some queries, then try and connect to the other with `DB::connection('spanner-two')`, it fails, with IAM permission denied errors.

It appears the auth from the first connection, which is now cached in `storage/framework/spanner/_auth`, is being loaded via the AuthCache into the second connection, and of course doesnt have the necessary permissions since the second connection should use an entirely different project/key file/auth

If however, I query the first connection, _manually delete_ the `storage/framework/spanner/` folder, then query the second one, it does work as expected.

I tried using `spanner:cooldown` on the first connection before attempting to use the second, but this didnt work, it doesnt appear to clear the `storage/framework/spanner/_auth` folder

taka-oyama:
Hi, this will be fixed in the next patch release. Thanks.

taka-oyama:
Fixed in `v8.1.0` and `v7.4.1`.

PR #210, opened 2024-05-20
fix: authCache needs namespace for each connection

taka-oyama:
Fix: #209

- [ ] backport to v7


halnique (approved):

taka-oyama:
Thanks for the review!

PR #211, opened 2024-06-02
Really basic attempt to get snapshots working

matthewjumpsoffbuildings:
Its kind of ugly, but if I just set Connection's `currentTransaction` property to a snapshot, it works.

I had to modify the type hint of the ManagesTransaction `currentTransaction` property, then just add a `snapshot()` method to Connection that accepts an array of snapshot init options, and a closure that runs while the snapshot is active

You can use it like so

```php
$conn->snapshot(['exactStaleness' => new Duration(15)], function() {
	return Model::all();
});
```

I am sure that you can improve on this significantly, but its a start

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-# v7.2.0 (Not released yet)
+# v7.3.0 (2024-04-17)
+
+Added
+- Request and Transaction tagging support (#206)
+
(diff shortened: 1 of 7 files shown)


matthewjumpsoffbuildings:
Sorry, I was on 7.3.0 since im still on Laravel 10. my mistake

Issue #218, opened 2024-06-07
phpstan fails on migrations when string('column', 'MAX') is used


oprudkyi:
spanner supports nonstandard `STRING(MAX)` column def  https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#scalars

in migrations it worked like 
```
$table->string('column', 'MAX');
```

while it works, phpstan fires error 
```
Parameter #2 $length of method Illuminate\Database\Schema\Blueprint::string() expects int|null, string given.
```

while  `@phpstan-ignore-next-line` helps probably some more advanced solution will be more usable

taka-oyama:
One potential way to resolve this is to interpret `-1` as `MAX` so it is compatible with phpdoc.

PR #223, opened 2024-06-24
fix: phpstan error for string('column', 'MAX')

taka-oyama:
Fix: #218 

## Checklist

- [x] CHANGELOG

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved a PHPStan error related to a type mismatch in the `string()` method of the `Blueprint` class.
	- Updated the method signature for `updateOrInsert` in version 8.1.2 for enhanced compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# v8.1.3 (2024-06-24)
+
+Fixed
+- phpstan error for string('column', 'MAX') (#223)
+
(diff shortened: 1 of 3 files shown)


zeriyoshi (approved):

zeriyoshi:
Instead of using stub, it may be better to use ignore. However, since this is a trivial matter, I will submit a separate PR for this later.

PR #227, opened 2024-09-06
feat: deprecate Connection::getDatabaseContext and rename to getMutationExecutor

taka-oyama:
The only ones using it now is the mutation trait, so just move it there.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method for query building, enhancing query capabilities.

- **Improvements**
  - Updated database operation methods for better mutation handling.
  - Optimized the `whereIn` method to improve query performance with large data sets.

- **Deprecations**
  - Deprecated the `getDatabaseContext()` method in favor of a new centralized mutation execution approach.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -128,7 +128,7 @@ class Connection extends BaseConnection
     }
 
     /**
-     * @deprecated will be removed in v9
+     * @deprecated will be removed in v10
      * @return Database|Transaction
      */
     protected function getDatabaseContext(): Database|Transaction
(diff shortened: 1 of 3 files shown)


src/Connection.php
```diff
@@ -128,6 +128,7 @@ public function getSpannerDatabase(): Database
     }
 
     /**
+     * @deprecated will be removed in v9
```
taka-oyama, line 131:
```suggestion
     * @deprecated will be removed in v10
```

halnique (approved):
LGTM

PR #229, opened 2024-09-20
feat: allow commit options to be set from config and Connection

### LINKED ISSUE #205: Add Support maxCommitDelay when insert and update
Hello

I noticed that Google recently released capacity-optimized recordings.

We can adjust the maximum writing delay, at the cost of increasing computational usage.

It would be very interesting if it were possible to define a default value in config.database or as a function during the querybuilder.

This can help with systems that need to work with less delay.

Source:
https://cloud.google.com/spanner/docs/throughput-optimized-writes

https://github.com/googleapis/google-cloud-php/blob/e22a5b45bd70c3cc27887d498e7a7d149de86b7b/Spanner/src/Operation.php#L127

taka-oyama:
Closes #205 

Requires the real Spanner to test this fully.

Feature request has been sent to emulator so it can be tested locally.
https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/184

## Summary by CodeRabbit

- **New Features**
	- Added support for snapshot queries to enhance query capabilities.
	- Introduced `whereNotInUnnest(...)` method in the query builder.
	- Added configurable commit options through a configuration file or method.

- **Improvements**
	- Updated `whereIn(...)` method to handle large value sets efficiently by using `UNNEST`.
	- Enhanced transaction management with new methods for getting and setting commit options.

- **Tests**
	- Added tests for committing transactions with specific options to ensure functionality.

### DIFF
--- a/tests/TransactionTest.php
+++ b/tests/TransactionTest.php
@@ -90,6 +90,38 @@ class TransactionTest extends TestCase
         Event::assertDispatchedTimes(TransactionRolledBack::class, 0);
     }
 
+    public function test_commit_with_options(): void
+    {
+        $conn = $this->getDefaultConnection();
+        /** @var Transaction $tx */
+        $tx = $conn->transaction(function (Connection $conn) {
+            return $conn->getCurrentTransaction();
+        });
+        $this->assertNotNull($tx);
+        dump($tx->getCommitStats());
+        $this->assertSame([], $tx->getCommitStats());
+        $this->assertSame([], $conn->getCommitOptions());
+
+        $newOptions = ['returnCommitStats' => true];
+        $conn->setCommitOptions($newOptions);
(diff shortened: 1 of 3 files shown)


tests/TransactionTest.php
```diff
@@ -90,6 +90,38 @@ public function testCommit(): void
...
 
+    public function test_commit_with_options(): void
+    {
+        $conn = $this->getDefaultConnection();
+        /** @var Transaction $tx */
+        $tx = $conn->transaction(function (Connection $conn) {
+            return $conn->getCurrentTransaction();
+        });
+        $this->assertNotNull($tx);
+        dump($tx->getCommitStats());
```
dkkoma, line 101:
```suggestion
```

PR #245, opened 2024-12-27
Support PHPStan 2.x

zeriyoshi:
PHPStan 2.0 Released 🎉
https://phpstan.org/

## Release Notes

- **Type Safety Improvements**
	- Enhanced type annotations across multiple classes and methods.
	- Refined parameter and return type definitions.
	- Improved static analysis support.

- **Dependency Updates**
	- Updated PHPStan from version 1 to version 2.
	- Stricter type checking configuration.

- **Configuration Changes**
	- Updated PHPStan configuration level to 9.
	- Removed previous error ignore settings.

- **Development Tooling**
	- Improved code quality through more precise type declarations.
	- Enhanced documentation for method signatures.

These changes primarily focus on improving type safety, code quality, and static analysis capabilities within the project.


src/Connection.php
```diff
@@ -589,8 +607,7 @@ protected function executeQuery(string $query, array $bindings, array $options):
 
         $tag = $this->getRequestTag();
         if ($tag !== null) {
-            $options['requestOptions'] ??= [];
-            $options['requestOptions']['requestTag'] = $tag;
+            $options['requestOptions'] ??= ['requestTag' => $tag];
```
taka-oyama, line 610: This will not add the tag if `requestOptions` alraedy exists.

src/Connection.php
```diff
@@ -108,6 +109,7 @@ public function __construct(
         $this->instanceId = $instanceId;
         $this->authCache = $authCache;
         $this->sessionPool = $sessionPool;
+        /** @phpstan-ignore argument.type */
```
zeriyoshi, line 112: Note: Rebase after #255 merged

src/Query/Concerns/UsesPartitionedDml.php
```diff
@@ -18,16 +18,18 @@
 namespace Colopl\Spanner\Query\Concerns;
 
 use Colopl\Spanner\Connection;
-use Colopl\Spanner\Query\Grammar;
+use Illuminate\Database\Query\Grammars\Grammar;
 
 /**
- * @property Connection $connection
+ * @template TConnection of Connection
```
taka-oyama, line 24: This is not necessary, we're not passing anything other than a Spanner's connection

src/Query/Grammar.php
```diff
@@ -189,7 +189,8 @@ protected function buildSearchFunction(string $function, array $where): string
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     * @return false
```
taka-oyama, line 193: Why tho?

src/Connection.php
```diff
@@ -234,7 +236,12 @@ protected function getDefaultPostProcessor(): QueryProcessor
     }
 
     /**
-     * @inheritDoc OVERRIDDEN for return type change
+     * OVERRIDDEN for return type change
+     *
+     * {@inheritDoc}
```
zeriyoshi, line 241: `{@inheritDoc}` is the strict syntax for inlining the parent PHPDoc. It is not required by many IDEs and utilities because of its ambiguous behavior, but to avoid ambiguity, it is used when overriding.
https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.inlineinheritdoc.pkg.html

src/Query/Grammar.php
```diff
@@ -142,39 +150,31 @@ protected function whereInUnnest(Builder $query, $where)
...
-    protected function whereSearchSubstring(Builder $query, array $where): string
+    protected function whereSearchSubstring(Builder $builder, array $where): string
     {
         return $this->buildSearchFunction('search_substring', $where);
     }
 
     /**
-     * @param string $function
      * @param array{ tokens: string, query: string, options: array<string, scalar> } $where
-     * @return string
```
taka-oyama, line 177: Please don't delete the return statements.

`@return string`

src/Connection.php
taka-oyama: Please submit this in a separate PR next time.
 
Add's too much noise to an already convoluted diff.

src/Query/Grammar.php
zeriyoshi: I wasn't aware of it (I changed it because I could show the shape more narrowly), but it was unnecessary. I will change it back.

src/Query/Concerns/UsesPartitionedDml.php
zeriyoshi: Indeed. I did not think deeply when generics support in `Builder` compatible. I will put it back.

src/Query/Grammar.php
```diff
@@ -110,24 +119,23 @@ protected function compileIndexHint(Builder $query, $indexHint)
     }
 
     /**
-     * @param Builder $query
      * @param array<string, Expression|string> $where
      * @return string
      */
-    protected function whereInArray(Builder $query, $where)
+    protected function whereInArray(Builder $builder, $where)
```
taka-oyama, line 125: If you look at the code in base class, all similar method names the Builder argument as `$query`. 
This should be reverted.

src/Concerns/ManagesTransactions.php
```diff
@@ -84,7 +84,7 @@ public function transaction(Closure $callback, $attempts = -1)
                     $this->transactions++;
 
                     $this->transactionsManager?->begin(
-                        $this->getName(), $this->transactions,
+                        $this->getName() ?? '', $this->transactions,
```
taka-oyama, line 87: This changes the behavior of the unerlying code.
Before -> error if null.
After -> no error.

This should be a separate PR and it should throw an error.

src/Query/Grammar.php
zeriyoshi: I made a mistake when I added `_` to the prefix of an unused parameter and then reverted it, as is my habit when using Psalm. I will change it back!

src/Concerns/ManagesTransactions.php
zeriyoshi: Initially I was throwing errors, but I stopped because the layers where the errors occur change.

taka-oyama: Fixing in another PR sounds good. Let's just add a ignore here for now.

src/Query/Grammar.php
zeriyoshi: If add type only return statement, PHPStan will warn me, so I want to either declare all of them or only the ones that need additional type information. Which would be better?

src/Schema/Blueprint.php
```diff
@@ -62,6 +62,7 @@ public function temporary()
 
     /**
      * @inheritDoc
```
zeriyoshi, line 64:
```suggestion
     * {@inheritDoc}
```

src/Query/Grammar.php
```diff
@@ -31,39 +31,48 @@ class Grammar extends BaseGrammar
...
      */
     public function compileInsertGetId(Builder $query, $values, $sequence)
     {
         $this->markAsNotSupported('insertGetId');
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     * @return ''
```
zeriyoshi, line 64:
```suggestion
     * @inheritDoc
```

src/Query/Grammar.php
zeriyoshi: In the meantime, I'll try to declare everything as it was originally.

src/Query/Processor.php
```diff
@@ -84,8 +102,9 @@ public function processColumns($results)
     }
 
     /**
-     * @param array $results
-     * @return array
+     * {@inheritDoc}
+     * @param array{index_name: string}&array<string, mixed> $results
```
zeriyoshi, line 106:
```suggestion
     * @param array{ index_name: string }&array<string, mixed> $results
```

src/Connection.php
taka-oyama: #255 has been merged.

src/Query/Grammar.php
taka-oyama: Yes. Just don't change formatting.

src/Query/ArrayValue.php
```diff
@@ -22,6 +22,9 @@
  */
 final readonly class ArrayValue
 {
+    /**
+     * @param array<array-key, mixed> $value
```
taka-oyama, line 26:
```suggestion
     * @param list<mixed> $value
```

zeriyoshi:
@taka-oyama
I fixed all review points.
Please check when you have time!

src/Query/Builder.php
```diff
@@ -23,11 +23,10 @@
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Query\Builder as BaseBuilder;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Collection;
 use LogicException;
 
 /**
- * @method Collection<int, array<array-key, mixed>> get($columns = ['*'])
+ * @template TConnection of Connection
```
taka-oyama, line 29: This generics is not necessary.

zeriyoshi: Without this Generics, type checking in the dependent in-house libraries will not work properly. This will be explained in the our company Slack.

src/Connection.php
```diff
@@ -304,28 +319,32 @@ public function cursorWithOptions(string $query, array $bindings, array $options
...
+     * @param array<array-key, mixed> $bindings
      */
     public function statement($query, $bindings = []): bool
     {
         // is SELECT query
         if (0 === stripos(ltrim($query), 'select')) {
-            return $this->select($query, $bindings) !== null;
+            $this->select($query, $bindings);
+            return true;
```
taka-oyama, line 330: This fix is unrelated to phpstan. Please submit it in a separate PR.

zeriyoshi: I have read the code to make sure that the change is correct, but it should not be false here, and PHPStan always warns as true, so I am making this change. Should I still add it to ignore?

src/Query/Builder.php
zeriyoshi: I found out that it probably wasn't a problem, so I removed it.

src/Connection.php
zeriyoshi: it means
- `$this->select()` method has language level return-type non-nullable `array`, `!== null` is always true.
- `$this->affectingStatement()` method has language level return-type non-nullable `int` , `!== null` is always true.

taka-oyama: Understood. 
It would have been nice if it was issued as a fix so it was easier to see what was going on.
Also would have made the review process slightly less painful for me.

taka-oyama (approved):

PR #246, opened 2025-01-06
Fix Grammar::substituteBindingsIntoRawSql fails during unnesting

zeriyoshi:
change #226 it possible for instances of `Colopl\Spanner\Query\Nested` objects to be included in the `bindings` of `Connection::getQueryLog()` when using UNNEST capable methods like `whereIn`. However, the Grammar was not compatible with this change. 

This PR fixes object handling.

## Summary by CodeRabbit

- **New Features**
	- Enhanced parameter handling in query methods
	- Improved support for nested data structures in database connections

- **Tests**
	- Added new test for SQL binding substitution
	- Introduced constant for parameter limit testing

- **Improvements**
	- Updated method signatures to support more flexible input types

### DIFF
--- a/src/Query/Grammar.php
+++ b/src/Query/Grammar.php
@@ -195,4 +195,17 @@ class Grammar extends BaseGrammar
     {
         return false;
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function substituteBindingsIntoRawSql($sql, $bindings)
+    {
+        $bindings = array_map(
+            static fn(mixed $binding): mixed => $binding instanceof Nested ? $binding->toArray() : $binding,
+            $bindings
+        );
+
+        return parent::substituteBindingsIntoRawSql($sql, $bindings);
+    }
 }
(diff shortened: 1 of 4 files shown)


src/Query/Grammar.php
```diff
@@ -195,4 +195,17 @@ public function supportsSavepoints()
     {
         return false;
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function substituteBindingsIntoRawSql($sql, $bindings)
```
taka-oyama, line 202: This is the wrong place to fix this. You should fix it in `Connection::escape()` https://github.com/colopl/laravel-spanner/blob/c21ad52052fc7ba79e088cb73bfcff41a5774f52/src/Connection.php#L450.

zeriyoshi: Sorry, fixed it

taka-oyama:
Test can be simplified to just testing the escape method.

taka-oyama (approved):

PR #260, opened 2025-01-17
feat: allow starting value to be set for generated columns

taka-oyama:
Previously you were only able to set starting value like 

```php
$table->integer('int_genv')->generatedAs('bit_reversed_positive start counter with 2');
```

Now it's possible to write it like 

```php
$table->integer('int_genv')->generatedAs()->startingValue(2);
```


halnique (approved):
LGTM

taka-oyama:
Thank you for reviewing.

Issue #231, opened 2024-10-01
Add snapshot methods to query builder/model


matthewjumpsoffbuildings:
Would be great if the snapshot method could be chained directly on query builder or model method calls. I tried to suss it myself to make a PR but I am not sure the best approach.

Ideally it would be great if you could do something like

```php
// snapshot from a model
User::snapshot()->all();

// snapshot from a builder
$queryBuilder
    ->where('foo', 'bar')
    ->snapshot()
    ->get();
```

Ideally you should be able to insert the `snapshot()` method anywhere in the chain, but since snapshots use a closure to execute under the hood Im not sure how it could be made chainable

taka-oyama:
A normal SELECT statement outside of a transaction is already a read only transaction.

matthewjumpsoffbuildings:
So running a SELECT sets `currentSnapshot` somewhere on the Connection and makes it execute the query in a snapshot? I am struggling to see where this happens in the code?

I remember there was some confusion between snapshots and timestamp bounds/single use transactions. Timestamp bounds and single use transactions are not the same as snapshots, as discussed here:

- https://github.com/colopl/laravel-spanner/issues/145#issuecomment-2143387161
- https://github.com/colopl/laravel-spanner/issues/145#issuecomment-2144377976
- https://github.com/colopl/laravel-spanner/issues/145#issuecomment-2144379350

matthewjumpsoffbuildings:
Also if it is the case that normal SELECT statements do somehow set `currentSnapshot`, how do you specify the timestamp bounds for the snapshot? Because the only place I see a snapshot being created is in the `ManagesSnapshots` trait, and the only time that happens is when `Connection::snapshot($timestampBound)` is called directly, which is surely not called when you just run a normal SELECT?

matthewjumpsoffbuildings:
I just tested this, if I run something like

```php
DB::connection()->table('totals')->get();
```

At no point does `Connection::executeSnapshotQuery` get called, nor does `Connection::snapshot`. Since `Connection::snapshot` seems to be the only place that actually runs `this->getSpannerDatabase()->snapshot($options);`, it is not the case that normal SELECT statements use snapshots. Rather, as before, they appear to be using single use transactions (optionally with timestamp bounds), which as discussed previously, are not the same as snapshots, and do not _guarantee_ lock free reads

matthewjumpsoffbuildings:
[...]

PR #267, opened 2025-03-02
Add snapshot methods to query Builder/Model

matthewjumpsoffbuildings:
As discussed in #231 this PR adds the ability to specify snapshot() when writing Eloquent queries on Models, or when using Builder directly

It can be used like so:

```php
// snapshot from a model
User::snapshot()->all();

// snapshot from a builder
DB::table('users')
    ->where('foo', 'bar')
    ->snapshot()
    ->get();
```

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -625,6 +625,10 @@ class Connection extends BaseConnection
             $options['requestOptions']['requestTag'] = $tag;
         }
 
+        if (isset($options['snapshotEnabled'])) {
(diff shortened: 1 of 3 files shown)


tests/SnapshotTest.php
```diff
@@ -46,6 +46,23 @@ public function test_snapshot(): void
         $this->assertSame('ok', $result);
     }
 
+    public function test_builder_snapshot(): void
```
taka-oyama, line 49: Please move this test to Query/BuilderTest.php

PR #274, opened 2025-04-30
fix!: SchemaBuilder::getIndexes was returning incorrect values

taka-oyama:
Was only returning the index name (also with wrong key name).
It should be returning more data.
This fix will align it with other database drivers.

Also, dropAllTables was probably not working correctly, so that was fixed as well.

### DIFF
--- a/src/Query/Processor.php
+++ b/src/Query/Processor.php
@@ -123,8 +123,8 @@ class Processor extends BaseProcessor
 
     /**
      * {@inheritDoc}
-     * @param list<array<string, mixed>> $results
-     * @return list<array{name: string, columns: list<string>, foreign_schema: string, foreign_table: string, foreign_columns: list<string>, on_update: string, on_delete: string}>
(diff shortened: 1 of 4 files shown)


src/Query/Processor.php
```diff
@@ -103,13 +103,21 @@ public function processColumns($results)
...
-            return ((object) $result)->index_name;
+            $result = (object) $result;
+
+            return [
+                'name' => $name = $result->name,
+                'columns' => $result->columns ? explode(',', $result->columns) : [],
+                'type' => strtolower($result->type),
+                'unique' => (bool) $result->unique,
+                'primary' => $name === 'PRIMARY_KEY',
+            ];
```
kt81, line 120: memo: Shape similar to https://github.com/laravel/framework/blob/12.x/src/Illuminate/Database/Query/Processors/MySqlProcessor.php#L86

kt81 (approved):
LGTM 🕺

PR #275, opened 2025-05-02
fix!: SchemaBuilder::getForeignKeys was returning incorrect values

taka-oyama:
Same reason as https://github.com/colopl/laravel-spanner/pull/274.


taka-oyama:
fixed broken test

halnique (approved):
LGTM

PR #273, opened 2025-04-28
fix: new phpstan errors

taka-oyama:
```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Schema/Builder.php                                                                                                                                                                                                                                                       
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  48     Return type (list<array{name: string, type: string, parent: string}>) of method Colopl\Spanner\Schema\Builder::getTables() should be compatible with return type (list<array{name: string, schema: string|null, schema_qualified_name: string, size: int|null, comment:  
         string|null, collation: string|null, engine: string|null}>) of method Illuminate\Database\Schema\Builder::getTables()                                                                                                                                                    
         🪪  method.childReturnType                                                                                                                                                                                                                                               
  75     Call to function in_array() with arguments string, list<array{name: string, columns: list<string>, type: string, unique: bool, primary: bool}> and true will always evaluate to false.  
         🪪  function.impossibleType                                                                                                                                                             
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your phpstan.neon.                                              
  150    Strict comparison using === between array{name: string, columns: list<string>, type: string, unique: bool, primary: bool} and 'PRIMARY_KEY' will always evaluate to false.              
         🪪  identical.alwaysFalse                                                                                                                                                               
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your phpstan.neon.                                              
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 


```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **New Features**
  - Added support for processing and standardizing table metadata, including schema and size details.
  - Enhanced table retrieval to allow optional schema filtering.

- **Improvements**
  - Table and column metadata now includes additional and more consistently formatted fields, including a new generation attribute for columns.
  - Table listing queries now support schema-based filtering for more precise results.

- **Documentation**
  - Improved and simplified internal documentation for better clarity.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #276, opened 2025-05-09
feat: allow auto increment to be set through Blueprint

taka-oyama:
#273 and #275 needs to be fixed for analyze to pass.


halnique (changes requested):
Shouldn't the following section of `README.md` be modified/updated/revised?

> Migrations
>
> Since Cloud Spanner does not support AUTO_INCREMENT attribute, Blueprint::increments (and all of its variants) will create a column of type STRING(36) DEFAULT (GENERATE_UUID()) to generate and fill the column with a UUID and flag it as a primary key.

taka-oyama:
Thank you for reminding about the README. It has been updated.

PR #279, opened 2025-05-09
feat: Blueprint should be injected via container

taka-oyama:


taka-oyama:
This replaces ALL blueprints. Which is not what we want. We only want to replace Spanner Connection's Blueprint.

PR #280, opened 2025-05-09
fix: SchemaBuilder::getColumns was missing values

taka-oyama:
The following is in this fix.

- added missing generation field.
- matched styling with other metadata getter methods. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accuracy of column metadata, including handling of nullable columns, default values, and auto-increment detection.
- **Tests**
  - Expanded test coverage to verify column details for nullable columns, columns with default values, and auto-increment columns.


src/Query/Processor.php
```diff
@@ -73,30 +73,25 @@ protected function processColumn(mixed $value): mixed
...
-     *     type_name: string,
-     *     type: string,
-     *     collation: null,
-     *     nullable: bool,
-     *     default: scalar,
-     *     auto_increment: false,
-     *     comment: null
-     * }>
+     * @param list<array<string, mixed>> $results
+     * @return list<array{name: string, type: string, type_name: string, nullable: bool, collation: string|null, default: mixed, auto_increment: bool, comment: string|null, generation: array{type: string, expression: string|null}|null}>
```
halnique, line 77: Why is the return type ambiguous for some attributes?
For example, `collation` is `string|null`, but it returns always `null`.

taka-oyama: It's because I copied it from laravel 12's phpdoc (which contains types).
This is now fixed.

src/Query/Processor.php
```diff
@@ -95,30 +95,25 @@ public function processTables($results)
...
-     *     type_name: string,
-     *     type: string,
-     *     collation: null,
-     *     nullable: bool,
-     *     default: scalar,
-     *     auto_increment: false,
-     *     comment: null
-     * }>
+     * @param list<array<string, mixed>> $results
+     * @return list<array{name: string, type: string, type_name: string, nullable: bool, collation: null, default: mixed, auto_increment: bool, comment: null, generation: null|null}>
```
halnique, line 99:
```suggestion
     * @return list<array{name: string, type: string, type_name: string, nullable: bool, collation: null, default: mixed, auto_increment: false, comment: null, generation: null}>
```

halnique (approved):
LGTM

PR #283, opened 2025-06-11
feat: Allow alter change stream to alter tables

taka-oyama:
Altering tables was always supported in Spanner but was never implemented.

### DIFF
--- a/src/Schema/Grammar.php
+++ b/src/Schema/Grammar.php
@@ -420,16 +420,18 @@ class Grammar extends BaseGrammar
     /**
      * @param Blueprint $blueprint
      * @param ChangeStreamDefinition $command
-     * @return string
+     * @return list<string>
      */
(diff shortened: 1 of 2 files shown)


src/Schema/Grammar.php
```diff
@@ -420,16 +420,18 @@ public function compileCreateChangeStream(Blueprint $blueprint, ChangeStreamDefi
...
      * @param ChangeStreamDefinition $command
-     * @return string
+     * @return list<string>
      */
-    public function compileAlterChangeStream(Blueprint $blueprint, ChangeStreamDefinition $command): string
+    public function compileAlterChangeStream(Blueprint $blueprint, ChangeStreamDefinition $command): array
     {
         $parts = [];
-        $parts[] = "alter change stream {$this->wrap($command->stream)}";
+        if ($command->tables !== []) {
```
dkkoma, line 428: if you want to change to `set for all` then whether `$command->tables` is empty, ddl should be create.

dkkoma (approved):

PR #297, opened 2026-05-15
self update composer

taka-oyama:
Fix: #

## Checklist

- [ ] CHANGELOG

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -2,10 +2,10 @@ services:
   test:
     build:
       context: .
-    command: /bin/sh -c "composer install --no-progress && vendor/bin/phpstan --no-progress --memory-limit=-1 && composer test"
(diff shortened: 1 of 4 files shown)


compose.yaml
```diff
@@ -2,10 +2,10 @@ services:
   test:
     build:
       context: .
-    command: /bin/sh -c "composer install --no-progress && vendor/bin/phpstan --no-progress --memory-limit=-1 && composer test"
+    command: /bin/sh -c "composer self-update && composer install --no-progress && vendor/bin/phpstan --no-progress --memory-limit=-1 && composer test"
```
tyabu12, line 5: Since `make build` already uses the `--pull` option, I think that the latest `composer:2` image is pulled on every CI run.

https://github.com/colopl/laravel-spanner/blob/master/Makefile#L4

Do we still need to run composer self-update?

taka-oyama: This was added for safety just in case so my token does not get exposed when checking.
This will not be merged.

PR #298, opened 2026-07-03
feat: cloud spanner v2

### LINKED ISSUE #290: add support for google-cloud/spanner v2
https://github.com/googleapis/google-cloud-php-spanner/releases/tag/v2.0.0

taka-oyama:
Closes https://github.com/colopl/laravel-spanner/issues/290


taka-oyama:
Waiting for the following to be fixed.
- https://github.com/googleapis/google-cloud-php/issues/9292
- https://github.com/googleapis/google-cloud-php/issues/9316
- https://github.com/googleapis/google-cloud-php/issues/9331
- https://github.com/googleapis/google-cloud-php/issues/9336

src/Console/WarmupCommand.php
```diff
@@ -26,7 +26,6 @@
 class WarmupCommand extends Command
 {
     protected $signature = 'spanner:warmup {connections?* : The database connections to be warmed up}
-               {--refresh : Will clear all existing sessions first.}
                {--skip-on-error : Will skip the connection if error is thrown.}';
 
     protected $description = "Warmup Spanner's Session Pool.";
```
dkkoma: `refreshSession()` always creates a new session now, so how about:
```
protected $description = "Creates a new session for Spanner.";
```

tests/Query/DatetimeTest.php
```diff
@@ -25,12 +25,27 @@
 use Google\Cloud\Spanner\SpannerClient;
 use Google\Cloud\Spanner\Timestamp;
 use Illuminate\Support\Carbon;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 class DatetimeTest extends TestCase
 {
-    public function testTimezone(): void
+    public function test_session_created(): void
```
dkkoma: Probably leftover debug code? It creates the database without any DDL, which would break the other tests if it ran first.

tests/SessionNotFoundTest.php
```diff
@@ -1,174 +0,0 @@
...
-            $config,
-            null,
-            $sessionPool,
-        );
-
-        $this->setUpDatabaseOnce($conn);
-        return $conn;
-    }
-
-    public function test_session_not_found_handling(): void
```
dkkoma, line 55: `handleSessionNotFoundException()` was rewritten but nothing covers it now.
Multiplexed sessions can't be deleted, but swapping the cached session for a
non-existent one reproduces it. How about restoring `SessionNotFoundTest` with this?

```
private function invalidateSession(Connection $conn): void
{
    $deleted = new Session();
    $deleted->setName($conn->getSessionName() . 'deleted');
    $deleted->setMultiplexed(true);
    // SessionCache::isExpired() dereferences the create time, so it must be set
    $deleted->setCreateTime(new ProtobufTimestamp(['seconds' => time()]));

    $item = $this->sessionCache->getItem(array_keys($this->sessionCache->getValues())[0]);
    $item->set($deleted->serializeToString());
    $item->expiresAt(new DateTimeImmutable('+1 day'));
    $this->sessionCache->save($item);

    $conn->disconnect();
}
```

composer.json
```diff
@@ -1,25 +1,26 @@
 {
   "name": "colopl/laravel-spanner",
   "description": "Laravel database driver for Google Cloud Spanner",
+  "version": "11.0.0",
   "type": "library",
   "license": "Apache-2.0",
```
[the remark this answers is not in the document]
dkkoma: Is the `version` field intentional? v11.0.0 is already out, and Composer recommends omitting it and letting the tag decide.

src/Concerns/ManagesSession.php
```diff
@@ -16,22 +16,31 @@
  * limitations under the License.
  */
 
-namespace Colopl\Spanner\Tests\Queue;
+namespace Colopl\Spanner\Concerns;
 
-use Colopl\Spanner\Connection;
-use Colopl\Spanner\Tests\TestCase;
-use Illuminate\Support\Facades\DB;
+use Google\Cloud\Core\EmulatorTrait;
```
dkkoma:
```suggestion
```

src/Console/WarmupCommand.php
taka-oyama: fixed

tests/Query/DatetimeTest.php
taka-oyama: yes thank you. I have removed it

tests/SessionNotFoundTest.php
taka-oyama: This was not possible when the expiration happens during a transaction because disconnecting will cause the transaction to clear. 

However, I was able to reproduce the not found error by altering the name of the session using reflection.

https://github.com/colopl/laravel-spanner/pull/298/commits/f19e096ad29c3772d3d3132a2336ebdf53ae706e

src/Connection.php
```diff
@@ -754,10 +733,7 @@ protected function shouldRunAsBatchDml(string $query): bool
      */
     protected function handleSessionNotFoundException(Closure $callback): mixed
     {
-        $this->disconnect();
-        // Currently, there is no way for us to delete the session, so we have to delete the whole pool.
-        // This might affect parallel processes.
-        $this->clearSessionPool();
+        $this->refreshSession();
```
dkkoma, line 787: `SessionCache::refresh()` differs from `ensureValidSession()` in that it takes no lock and
always creates a new session. So once a `SessionNotFound` occurs, a single server can end
up using several sessions at the same time.

With a shared-nothing setup like PHP-FPM, this only affects the requests that are in flight
at that moment, since every request reads the session cache from scratch.

With resident workers like FrankenPHP, each process/thread keeps its own `SessionCache`,
so every one of them will eventually call `refresh()` and create its own session.

This may need a change on the upstream library side — a supported way to invalidate the
cached session, so that it is re-created under the same lock `ensureValidSession()`
already uses.

taka-oyama: I noticed that `SessionCache::name()` calls `ensureValidSession()` every time it's called.
Since `SessionCache::name()` is called on all operation like `Operation::execute()` and `Operation::commit()`, right before the operation is executed, sessions are guaranteed to renew.

So the only time this will every trigger is when the server side deletes the sessions for whatever reason.

dkkoma: You're right, `name()` runs before every operation, so the rotation can't be missed.
Thanks for digging into it.

src/Connection.php
```diff
@@ -651,8 +626,12 @@ protected function executeQuery(string $query, array $bindings, array $options):
...
+        $snapshotOptions = isset($options['databaseRole'])
+            ? ['databaseRole' => $options['databaseRole']]
+            : [];
+
         $snapshot = $this->getSpannerClient()
-            ->batch($this->instanceId, $this->database, $options)
+            ->batch($this->instanceId, $this->database, $snapshotOptions)
             ->snapshot();
 
         foreach ($snapshot->partitionQuery($query, $options) as $partition) {
```
halnique: When using Data Boost, `partitionQuery()` throws an exception if `databaseRole` or staleness options are present in the query.

In `google/cloud-spanner` v2, `Operation::partitionQuery()` enforces strict option validation. 
Currently, `databaseRole` gets copied into `$snapshotOptions` but is never removed from the original `$options` array. When this unfiltered array is passed to `partitionQuery()`, `OptionsValidator` rejects the leftover keys and the query fails before hitting the server.

```php
$conn->selectWithOptions('SELECT * FROM User', [], ['dataBoostEnabled' => true, 'databaseRole' => 'x']);
// QueryException: Unexpected option(s) provided: databaseRole
```

Using `Arr::pull($options, 'databaseRole')` instead of just reading the key fixes this specific case.

However, the query builder suffers from the same pass-through bug. `executeQuery()` handles `dataBoostEnabled` before consuming snapshot/staleness keys. Chaining `useDataBoost()` with `withStaleness()` or `snapshot()` passes those leftover keys down, triggering the exact same strict validation error.

```php
$conn->table('User')->useDataBoost()->withStaleness(new ExactStaleness(10))->get();
// Unexpected option(s) provided: exactStaleness

$conn->table('User')->useDataBoost()->snapshot(new ExactStaleness(10))->get();
// Unexpected option(s) provided: snapshotEnabled, snapshotTimestampBound
```

(Note: Verified on the emulator with `google/cloud-spanner v2.10.5.` In v1, these extra keys were silently ignored so staleness didn't actually apply, but v2 turns this into a hard runtime error).

Suggested Fix

Filter/Route keys: Extract `databaseRole` into the `batch()` options (e.g., using `Arr::pull()`). Forward timestamp bounds explicitly via `snapshot(['transactionOptions' => ...])` so stale reads actually work, or reject the combination with a clear, custom error.

Add Tests: We currently only have `test_dataBoost_enabled`. We should add test coverage for `useDataBoost()` combined with `databaseRole`, `withStaleness()`, and `snapshot()` to catch this class of issues.

halnique (commented):
The README isn't touched by this PR, but a good part of it documents exactly what this PR removes, so anyone upgrading will be following instructions that no longer work.

src/Connection.php
taka-oyama: exactStaleness on data boost should work now.

Also improved the error message for other options as well.

taka-oyama:
Global timeout was not working so that has been readded with [c2cf6ad](https://github.com/colopl/laravel-spanner/pull/298/commits/c2cf6ad9bb51b0a814aca146a4514b5ef2802017).

README.md
```diff
@@ -446,24 +441,15 @@ All mutations calls within a transaction are queued and sent as batch at the tim
...
-of your server is recommended. This can be achieved by running the `php artisan spanner:cooldown` command.
+### SessionCache and AuthCache
 
-### Queue Worker
+In order to improve the performance of the first connection per request, we use [AuthCache](https://github.com/googleapis/google-cloud-php#caching-access-tokens) and [SessionCache](https://docs.cloud.google.com/spanner/docs/sessions#multiplexed_sessions).
 
-After every job is processed, the connection will be disconnected so the session can be released into the session pool. 
-This allows the session to be renewed (through `maintainSessionPool()`) or expire.
+By default, this library uses [Filesystem Cache Adapter](https://symfony.com/doc/current/components/cache/adapters/filesystem_adapter.html) for caching sessions. 
+If you want to use a different adapter, you can extend ServiceProvider and inject it into the constructor of `Colopl\Spanner\Connection`.
```
halnique, line 449: Thanks for reworking this section — the `session_pool` block, `maxSessions`, the "60 minutes" note with `spanner:cooldown`, and the `### Queue Worker` section are all gone now. Three leftovers from v1 remain, and the one on this line is the one that fails silently.

**These lines (L448-449).** The extension point they describe was renamed *and* retyped in this PR:

- `SpannerServiceProvider::createSessionPool(): SessionPoolInterface` → `createSessionCache(): AdapterInterface`
- `Connection`'s 6th constructor argument: `?SessionPoolInterface $sessionPool` → `?CacheItemPoolInterface $sessionCache`

Anyone who followed this exact instruction on v1 has an override of `createSessionPool()`. After upgrading, the parent never calls it again — no error, no deprecation warning — and their custom adapter silently falls back to the default `FilesystemAdapter`. Could this paragraph name `createSessionCache()` explicitly so the rename is discoverable from the docs?

Two more spots outside this diff hunk, so I couldn't anchor them separately:

- **L53**: "You can pass `SpannerClient` config and `CacheSessionPool` options as below." `Google\Cloud\Spanner\Session\CacheSessionPool` no longer exists in v2, and the `session_pool` block was removed from the example directly below — so the sentence and the example now contradict each other.
- **L241**: "This creates a new session in the background which is not shared with the current session pool." There is no pool anymore (wording only).

Minor, same section: L451 says "warming up **the sessions**" (plural), but `spanner:warmup` now refreshes a single session.

taka-oyama: README.md has been updated and UPGRADE.md has been added to inform about changes in detail.

tests/SessionNotFoundTest.php
```diff
@@ -19,53 +19,34 @@
...
-        $this->assertSame([12345], $conn->selectOne('SELECT 12345'));
+    private function deleteSession(Connection $conn): void
+    {
+        $sessionCache = $conn->getSpannerDatabase()->session();
+        $refClass = new \ReflectionClass($sessionCache);
+        $sessionProp = $refClass->getProperty('session');
+        $session = $sessionProp->getValue($sessionCache);
+        $sessionClass = new \ReflectionClass($session);
+        $nameProp = $sessionClass->getProperty('name');
+        $nameProp->setValue($session, $session->getName() . '_deleted');
```
halnique, line 54: `deleteSession()` rewrites `Session::name` on the `SessionCache` instance the connection currently holds, leaving the shared cache entry pointing at a live session. In production the cached entry is the dead one. `reconnect()` therefore re-reads a valid session from the cache, and these tests pass with `$this->refreshSession()` removed from `handleSessionNotFoundException()`. Measured on the emulator:

| `deleteSession()` | with `refreshSession()` | without |
|---|---|---|
| as written today | 6 passed | **6 passed** |
| also poisoning the shared cache entry | 6 passed | **6 errors** |

Poisoning the cache entry needs no additional `SessionCache` internals. If the test injects its own pool, as `test_session_cache()` does, it can write to it directly:

```php
$item = $this->sessionPool->getItem(array_key_first($this->sessionPool->getValues()));
$cached = new Session();
$cached->mergeFromString($item->get());
$cached->setName($cached->getName() . '_deleted');
$item->set($cached->serializeToString());
$this->sessionPool->save($item);
```

composer.json
```diff
@@ -12,15 +12,15 @@
     "ext-grpc": "*",
     "ext-json": "*",
     "laravel/framework": "^13.20",
-    "google/cloud-spanner": "^1.106.0",
+    "google/cloud-spanner": "^2.10",
```
halnique: `^2.10` admits `v2.10.0`, which predates all four fixes; the last of them landed in `v2.10.5`. It stays resolvable for anyone who already has it locked. Pinned to it, on a clean emulator:

| | v2.10.0 | v2.10.6 |
|---|---|---|
| suite | **2 errors** | 247 passed |
| duration | 12m23s | 48s |

Both errors are `ALREADY_EXISTS: Row {...} already exists` from `Transaction::commit()` (#9331), in `testInsertUsingMutationWithTransaction` and `testDeleteUsingMutationWithDifferentArgs`. Each test generates its own `uuid4()`, so this is one test inserting twice, not leftover data. The 15x duration is #9316.

`^2.10.5` covers all four.

taka-oyama:
```suggestion
    "google/cloud-spanner": "^2.10.5",
```

tests/SessionNotFoundTest.php
taka-oyama: This has been addressed in https://github.com/colopl/laravel-spanner/pull/298/commits/957d50fd5c88c28a7f7af5087421c8f9feceeec5

src/Concerns/ManagesSession.php
```diff
@@ -16,15 +16,30 @@
...
+        $this->getSpannerDatabase()->session()->refresh();
+    }
+
+    /**
+     * @return string
+     */
+    public function getSessionName(): string
     {
-        DB::connection()->select('SELECT 1');
+        return $this->getSpannerDatabase()->session()->name();
```
halnique, line 43: Is `withDefaultTimeout()` not needed here?
I would like you to review all calls outside of transactions to make sure there are no other places where the behavior has changed.

tests/TransactionTest.php
halnique: Shouldn't we fix this part as well, just like L102?

src/Concerns/ManagesSession.php
taka-oyama: I don't understand what's being asked here. 
Nothing in getSessionName allows timeout options?

The diff here is kinda weird, it's replacing `Colopl\Spanner\Tests\Queue\QueryJob` with `Colopl\Spanner\Concerns\ManagesSession` so maybe there is some confusion here...?

tests/TransactionTest.php
taka-oyama: Thank you, this had been addressed.

src/Concerns/ManagesSession.php
halnique: Sorry, wrong place! I'll re-post this in the right thread.

taka-oyama: I think I know where you're referring to. Mutation methods did not have default timeouts. I will add them now.

src/Concerns/ManagesTransactions.php
halnique: It would probably be better to pass options, including a timeout, to `Transaction::rollback()` as well.

composer.json
```diff
@@ -12,15 +12,15 @@
...
+    "google/cloud-spanner": "^2.10.5",
     "grpc/grpc": "^1.42",
     "symfony/cache": "~7 || ~8",
     "symfony/lock": "~7 || ~8"
   },
   "require-dev": {
     "friendsofphp/php-cs-fixer": "^3.95",
     "orchestra/testbench": "~11",
-    "phpunit/phpunit": "~12",
+    "phpunit/phpunit": "~11.0",
```
halnique: Is this downgrade intentional?

taka-oyama: Oops, that was supposed to be a temporary fix.
This was lowered because 12.x was raising the following notices.

```
* No expectations were configured for the mock object for Google\Cloud\Spanner\Database. Consider refactoring your test code to use a test stub instead. The #[AllowMockObjectsWithoutExpectations] attribute can be used to opt out of this check.
```

This is now fixed in [057ee43](https://github.com/colopl/laravel-spanner/pull/298/commits/057ee431bb73481898ae0175d37cc61dc96bf39c).

src/Concerns/ManagesTransactions.php
taka-oyama: I've scanned through all the code and indeed there were many places where timeout was missing.
I think the default timeout needs to be revived.
I've made a PR upstream to see what the authors of `google/cloud-spanner` think.
https://github.com/googleapis/google-cloud-php/issues/9617

halnique (approved):

src/Connection.php
```diff
@@ -651,11 +650,29 @@ protected function executeQuery(string $query, array $bindings, array $options):
...
+        $snapshotOptions = [
+            'transactionOptions' => $this->extractOptions($options, ['strong', 'readTimestamp', 'exactStaleness']),
+        ];
+        $partitionOptions = $this->extractOptions($options, [
+            'maxPartitions',
+            'partitionSizeBytes',
+            'parameters',
+            'types',
+            'dataBoostEnabled',
+            'timeoutMillis',
```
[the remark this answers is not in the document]
taka-oyama: executePartition does not take second arguments.

PR #299, opened 2026-07-15
feat: laravel 13

### LINKED ISSUE #295: laravel 13 support

taka-oyama:
Closes https://github.com/colopl/laravel-spanner/issues/295

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:8.4-cli-alpine
+FROM php:8.3-cli-alpine
 
 COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
 
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ Laravel database driver for Google Cloud Spanner
 
 ## Requirements
 
-- PHP >= 8.2
-- Laravel >= 11
+- PHP >= 8.3
+- Laravel >= 13
 - [gRPC extension](https://cloud.google.com/php/grpc)
 - [protobuf extension](https://cloud.google.com/php/grpc#install_the_protobuf_runtime_library) (recommended for better performance)
 - `sysvmsg`, `sysvsem`, `sysvshm` extensions (recommended for better performance)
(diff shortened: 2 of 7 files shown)


halnique (commented):
`README.md` L13 still says `Laravel >= 11`, but this bump makes 11/12 unresolvable. Worth updating in the same commit:

```diff
- - Laravel >= 11
+ - Laravel >= 13
```

taka-oyama:
@halnique Thank you. README.md has been updated.

dkkoma:
Looks like `fetchUsing` should also be rejected in `Builder::runSelect()`.

dkkoma:
PHPUnit 13 requires PHP 8.4.1+, so composer.json should be updated accordingly.

taka-oyama:
> Looks like fetchUsing should also be rejected in Builder::runSelect().

`Builder::runSelect(...)` does not take any arguments...?

> PHPUnit 13 requires PHP 8.4.1+, so composer.json should be updated accordingly.

PHPUnit 13 is a dev only requirement. The requirement for the package itself should still be 8.3+.

taka-oyama:
I've lowered the phpunit version to 12 and lowered the docker requirement so it can run on 8.3 instead of 8.4

dkkoma:
> `Builder::runSelect(...)` does not take any arguments...?

Sorry, I should have been clearer.
In Laravel 13, `fetchUsing(...)` stores the value in the `Builder::$fetchUsing` property, and the base `runSelect()` forwards it to `Connection::select()`.

However, this package's `Query\Builder::runSelect()` override calls `$this->connection->selectWithOptions(...)` directly, so the guard added to `Connection::select()` never runs on this path. `User::query()->fetchUsing(...)->get()` silently ignores fetchUsing instead of throwing, the exact behavior this PR is trying to prevent.

So the suggestion is to check the property, e.g.:

```
 protected function runSelect()
 {
     if ($this->fetchUsing !== []) {
         throw new InvalidArgumentException('$fetchUsing is not supported by Cloud Spanner');
     }
     // ...
 }
```

taka-oyama:
Should be addressed now.

dkkoma (approved):

PR #303, opened 2026-09-02
fix: incorrect index name when index specifies order

taka-oyama:
This pull request improves the way index names are generated in the schema blueprint, ensuring consistent naming regardless of whether columns are specified as a list or an associative array. It also updates the corresponding tests to cover these scenarios.

Index name generation improvements:

* Modified `createIndexName` in `Blueprint.php` to use the correct column names when columns are provided as an associative array (e.g., with sort directions), ensuring index names are generated consistently.

Test coverage enhancements:

* Updated `BlueprintTest.php` to add test cases for creating indexes with both simple column lists and associative arrays specifying sort directions, verifying the new index name logic.

### DIFF
--- a/src/Schema/Blueprint.php
+++ b/src/Schema/Blueprint.php
@@ -489,7 +489,7 @@ class Blueprint extends BaseBlueprint
 
     /**
      * @inheritDoc
-     * @param list<string> $columns
+     * @param array<string, string>|list<string> $columns
      */
     protected function createIndexName($type, array $columns)
     {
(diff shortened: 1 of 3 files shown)


src/Schema/Blueprint.php
tyabu12:
```suggestion
     * @param string|array<string, string>|list<string> $columns
```

halnique (approved):

tyabu12 (approved):
LGTM!

PR #304, opened 2026-09-10
feat: Clear session when dropping a database

taka-oyama:
The old session will throw an error when attempting to use it after dropping a database so it's best to clear it when the database is cleared.

### DIFF
--- a/src/Concerns/ManagesDataDefinitions.php
+++ b/src/Concerns/ManagesDataDefinitions.php
@@ -20,6 +20,7 @@ namespace Colopl\Spanner\Concerns;
 
 use Google\Cloud\Core\LongRunning\LongRunningOperation;
 use Google\Cloud\Spanner\Database;
+use Psr\Cache\CacheItemPoolInterface;
 use RuntimeException;
 use function json_encode;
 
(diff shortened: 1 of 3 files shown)


src/Concerns/ManagesDataDefinitions.php
```diff
@@ -30,6 +31,11 @@ trait ManagesDataDefinitions
      */
     abstract public function getSpannerDatabase(): Database;
 
+    /**
+     * @return CacheItemPoolInterface|null
+     */
+    abstract protected function getSessionCache(): ?CacheItemPoolInterface;
```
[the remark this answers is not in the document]
taka-oyama: this is a new major version update. it's fine.

t-i-job:
一通り見ましたが、もう1人ぐらい見てもらうという話をMTGでしました。

src/Connection.php
```diff
@@ -136,6 +136,14 @@ public function getSpannerDatabase(): Database
         return $this->spannerDatabase ?? throw new LogicException('Spanner Database does not exist');
     }
 
+    /**
+     * @return CacheItemPoolInterface|null
+     */
+    protected function getSessionCache(): ?CacheItemPoolInterface
+    {
+        return $this->sessionCache;
```
halnique, line 155: If `config['client']['cacheItemPool']` is specified, it will be used preferentially, so there is a possibility that it may diverge from this return value.

taka-oyama: This has been addressed in [20ce81e](https://github.com/colopl/laravel-spanner/pull/304/commits/20ce81e64cfc4db45431dca3a15e6fe944f9706c)

halnique (approved):

